PR #4, opened 2017-12-13
[oracles] Initial notes on oracles

gorillainduction:


oracles/oracle_life_cycle.md
```diff
@@ -0,0 +1,76 @@
...
+When the response transaction has been accepted to the chain, Bob's
+event subscription notifies him that there is a response.
+
+The event contains:
+- The location of the transaction (?).
+- The content of the response.
+- The TTL of the response.
+
+Bob can now use the response by:
+- Executing a smart contract which references the response.
```
UlfNorell: The smart contract needs to reference the *query*, not the response, since (typically) you would publish the contract before you get the response.

gorillainduction: Agreed. It will probably be the same Id, though.

PR #6, opened 2017-12-13
Discuss ttl

gorillainduction:


knarz:
The oracle will have to require enough fees to cover operation and transaction fees.

PR #8, opened 2018-01-08
first draft for the AENS

knarz:
This is the draft for the first version of the AENS.

Comments and questions are welcome.

### DIFF
--- a/drafts/AENS.md
+++ b/drafts/AENS.md
@@ -112,7 +112,11 @@ into the `revoked` state. This value MUST NOT be further than
 
 ***ttl***: a suggestion as to how long any clients should
 cache this information. (***TODO***: should have a reasonable
-upper limit, e.g. 86400 seconds)
+upper limit, e.g. 86400 seconds, and probably a different
+name to not be confused with the general TTL for transactions)
+
+This entry is only relevant for clients and has no conensus
+impact.
 
 ***pointers***: a dictionary with all the values this entry
 points towards, e.g. `{kind: "ipfs", data: "QmVcSqVEsvm5RR9mBLjwpb2XjFVn5bPdPL69mL8PH45pPC"}`.
@@ -244,8 +248,8 @@ This transaction starts the claiming process.
  ------------ ----
 ```
 
-A `pre-claim` transaction for a name that is in the `claimed`
-state should be considered invalid.
+A client interacting with the blockchain should generate a warning
+if a user tries to claim a name that is not available.
 
 The `pre-claim` has an implicit expiration attached to it. A
 `pre-claim` MUST be considered invalid after 300 blocks, i.e.
@@ -304,8 +308,8 @@ Note that the fee here is a distinct fee from the normal transaction fee.
 The `update` transaction MUST be signed by the owner
 of the name entry to be updated.
 
-The `expire_by` MUST NOT be more than 1 year into the
-future.
+The `expire_by` MUST NOT be more than 36000 blocks into
+the future.
 
 
 #### Transfer
@@ -439,5 +443,6 @@ be distributed to random accounts via a lottery.
 [1] Kalodner, Harry A., et al. "An Empirical Study of Namecoin and Lessons for Decentralized Namespace Design." WEIS. 2015.
 
 [2] Ali, Muneeb, et al. "Blockstack: A Global Naming and Storage System Secured by Blockchains." USENIX Annual Technical Conference. 2016.
+
 
 
--- a/drafts/AENS.md
+++ b/drafts/AENS.md
@@ -20,16 +20,16 @@ humans trying to use these systems.
 A solution to a similar problem is the Domain Name System, or DNS, which
 should be familiar to most people, providing human readable names for IP
 addresses. The issue with DNS is its centralization and it was conjectured
-by [Zooko](https://en.wikipedia.org/wiki/Zooko%27s_triangle) that a 
+by [Zooko](https://en.wikipedia.org/wiki/Zooko%27s_triangle) that a
 decentralized system can't have human-meaningful names while still being
 secure. This conjecture has been proven to be false by a number of systems,
 one example being [NameCoin](https://www.namecoin.org), which was debuted
 in 2011. It allowed users to associate names with a number of key-value
-pairs. The NameCoin network still exists but sees barely any usage.
+pairs. The NameCoin network still exists today but sees barely any usage.
 
-A more recent effort is ENS, the Ethereum Name Service, which enables users 
+A more recent effort is ENS, the Ethereum Name Service, which enables users
 to claim names via a set of smart contracts. ENS is currently running a two
-year trial deployment, during which users can claim name with a `.eth` 
+year trial deployment, during which users can claim name with a `.eth`
 postfix, e.g. `mywallet.eth`. ENS currently allows users to associate a single
 32 byte array of data with a node, typically used for a hash.
 
@@ -45,7 +45,7 @@ or `привет.aet` are part of the same `.aet` namespace.
 This first draft of the AENS is not going to have any governance mechanism but
 will only allow registrations under a single namespace: `.aet`.
 We will also not include mechanisms similar to the `sunrise` and `landrush`
-periods, which are common for DNS, where registration is restricted to small 
+periods, which are common for DNS, where registration is restricted to small
 select groups and trademark holders in order to avoid name squatting, before
 the general public can start claiming names.
 
@@ -56,25 +56,18 @@ It is unclear what a good mechanism for a naming system would look
 like. If we imagine two actors both being interested in the same name,
 what would a »fair« solution be to resolve this?
 
-Fees are the main mechanism to discourage spam and squatting.
+Fees are the main mechanism to discourage spam and squatting. This
+initial version will burn the governance fee, in order to enable us
+to change this behaviour in the future. If we start out by giving this
+fee to miners, they will be very hesitant to accept any changes, which
+will impact their income negatively. Thus starting out with burning
+could allow us an easier path for future update to the fee structures.
 
-***Options:***
+Every entry in the `.aet` namespace pays the same amount of fees.
 
-1. one time fee, which can be claimed by miner
-2. one time fee which gets burned
-3. deposit which can be reclaimed after the name is released
-4. no additional fee other than the standard transaction fee
-
-For the actual amount of fees to be paid we have a couple of options:
-
-***Options:***
-
-1. based on length (e.g. have buckets of different price categories)
-2. every name pays the same amount of fees (or none)
-
-Each entry has a fixed expiration date on claim after which the entry should 
-transition into the `revoked` state. Once it reaches this state, the name 
-will be released, i.e. transition into the `unclaimed` state, after which 
+Each entry has a fixed expiration date on claim after which the entry should
+transition into the `revoked` state. Once it reaches this state, the name
+will be released, i.e. transition into the `unclaimed` state, after which
 it could be claimed again.
 
 To prevent the entry from expiring a user can, at any point before reaching
@@ -83,18 +76,20 @@ into the future by the time delta of the registration time and the update
 time, e.g. if a user posts an update one week after claiming their name,
 the expiration date gets pushed one week further into the future.
 The main motivation of this expiration date is to prevent situations where
-the private keys, which control the name, are lost making the name unusable 
+the private keys, which control the name, are lost making the name unusable
 as well.
 
 
 ## Specification
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", 
-"SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be 
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be
 interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 
 ### AENS Entry
+
+This is the entry as it should be stored by a node.
 
 ```
   name         size (bytes)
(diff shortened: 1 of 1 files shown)


drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+upper limit, e.g. 86400 seconds)
+
+***pointers***: a tuple with all the values this entry
+points towards, e.g. `{kind: "ipfs", data: "QmVcSqVEsvm5RR9mBLjwpb2XjFVn5bPdPL69mL8PH45pPC"}`
+
+
+### Name
+
+Names MUST be normalized according to IDNA2008 before hashing, as 
+described in [uts-46](https://unicode.org/reports/tr46) and SHOULD
```
happi: To fully specify this, what should the parameter values for the following parameters be?
A boolean flag: UseSTD3ASCIIRules
A boolean flag: CheckHyphens
A boolean flag: CheckBidi
A boolean flag: CheckJoiners
A processing_option with one of two values:
Transitional_Processing for transitional handling of Deviation characters, or
Nontransitional_Processing otherwise

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+
+We are going to store the AENS entries in an ESMT and use the
+hash of a name, as defined above, as a pointer to the entry data.
+If there's no entry for a given hash then that pointer will point
+to an empty hash or to the hash of that entry data otherwise.
+
+
+## Launch
+
+Giving users plenty time to prepare for a launch of the naming
```
happi: Is the .test namespace intended to be available as soon as the feature is released and other top level domains only available after launch?
This would make it possible to write clients and test the feature before the launch.

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+
+1. (optional) wait `n` blocks, s.t. that the block including the `pre-claim` cannot be reversed whp 
+2. send `claim` transaction to reveal name and pay the associated fee
+
+If the time delta of 1 and 2 is bigger than e.g. 48 hours then the 
+`pre-claim` expires again.
+
+The `claim` transaction MUST be signed by the same private key as the 
+`pre-claim` transaction.
+
```
happi: What happens if the name is already claimed?
Will the claim transaction be considered invalid an not included, hence no fee, or will it be consider just failing and still costing the fee?

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+| name       | 63 |
+ ------------ ----
+| fee        | 32 |
+ ------------ ----
+```
+
+Flow for a user:
+
+1. (optional) wait `n` blocks, s.t. that the block including the `pre-claim` cannot be reversed whp 
+2. send `claim` transaction to reveal name and pay the associated fee
```
happi: What is the relation between multiple pre-claims and the claim?
What stops a miner from posting his own pre claim and dropping the claim transaction when he sees the claim transaction? Then posting his own claim transaction?

happi: I guess that if an older (but live) pre claim transaction exists then the claim could go to the first pre claim this would require the malicious miner to mine all blocks till the first claim expires.

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+
+### Namespaces/Labels
+
+A name is split up into labels by the `FULL STOP (U+002E .)` character.
+These labels will also be referred to as namespaces, e.g. `mywallet.aet` 
+can be understood as the `mywallet` namespace in the `aet` namespace.
+
+A label MUST be longer than three characters.
+
+Labels SHOULD be at most 63 characters long and the full path
```
thepiwo: so a name with one label and the default namespace has a maximum of 63 + 1 + 3 characters?

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+long, in order to enable interoperability with DNS.
+
+
+### Registrars
+
+A registrar controls who is allowed to claim names in their namespace
+and under what circumstances they are allowed to do so.
+
+The only available registrars for this AENS version will be hard-coded
+ones, which own the `.`, `.aet` and `.test` namespaces.
```
thepiwo: why do we want to have a `.test` namespace in the mainnet? I'd prefer having `.aet` in the mainnet and `.test` in the testnet

drafts/AENS.md
thepiwo: I think a miner could only cheat by not including the original pre-claim and replacing it with a new one. This can be worked arround by including the account in the namehash.

**JUST AS IDEA**

```
def namehash(name, account):
  if name == '':
    return '\0' * 32
  else:
    label, _, remainder = name.partition('.')
    return sha256(namehash(remainder, account) + sha256(label) + account)
```
(not optimal as account is hashed for each remainder)

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+[EIP-137](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-137.md#namehash-algorithm)
+using SHA256 instead.
+
+```
+def namehash(name):
+  if name == '':
+    return '\0' * 32
+  else:
+    label, _, remainder = name.partition('.')
+    return sha256(namehash(remainder) + sha256(label))
```
thepiwo: I think of the names as `subsub.sub.label.tld` wouldn't this mean that we should work recursively on the label? why is the recursion used anyway? 

`return sha256(namehash(label) + sha256(remainder))`

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+`pre-claim` transaction.
+
+
+#### Update
+
+```
+ ------------ ----
+| hash       | 32 |
+ ------------ ----
+| expire_by  | 8  |
```
thepiwo: Would this only expire the update itself not the claimed name?

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+
+#### Claim
+
+```
+ ------------ ----
+| name       | 63 |
+ ------------ ----
+| fee        | 32 |
+ ------------ ----
+```
```
thepiwo: does the claim have an expiration as well? otherwise can't it be claimed forever and not be updated ever, to block the name?

Can't we combine the `claim` and `update` so a first claim would include the initial data and an update for renewal purposes would just claim the same name for the same owner again to push back the expiry

drafts/AENS.md
knarz: Yes, the `.test` namespace should be available on the testnet, as soon as possible.

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+| ttl        | 8  |
+ ------------ ----
+| pointers   |    |
+ ------------ ----
+```
+
+***owner***: the account, which controls this entry.
+
+***expires_by***: the blockheight after which the entry goes
+into the `revoked` state.
```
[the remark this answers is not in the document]
knarz: Hey, a bit further down it says that the expiration date must be at least a year in the future but I'll add it here as well.

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+
+***ttl***: a suggestion as to how long any clients should 
+cache this information. (***TODO***: should have a reasonable 
+upper limit, e.g. 86400 seconds)
+
+***pointers***: a tuple with all the values this entry
+points towards, e.g. `{kind: "ipfs", data: "QmVcSqVEsvm5RR9mBLjwpb2XjFVn5bPdPL69mL8PH45pPC"}`
```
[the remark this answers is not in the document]
knarz: Ideally I'd avoid json and instead use one of the serialisation formats that we settle on. Making this data structure too complex might also impact clients trying to read these values. But I generally agree.

drafts/AENS.md
knarz: Yes

drafts/AENS.md
knarz: This can be used to update the `expire_by` date. The update itself has no expiration.

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+
+Fees are the main mechanism to discourage spam and squatting.
+
+***Options:***
+
+1. one time fee, which can be claimed by miner
+2. one time fee which gets burned
+3. deposit which can be reclaimed after the name is released
+4. no additional fee other than the standard transaction fee
+
```
emil-apeunit: 3. is quite weak to prevent squatting and spam, 2. seems to be the safest option to prevent fraud for now and the most useful option might be to collect them and use the governance system to invest them into the Eco-system, for example: improve the naming system or other parts of the blockchain.

drafts/AENS.md
knarz: If a hash/name is already claimed then the transaction can be considered invalid. Ideally a wallet would warn a user trying to claim such a name.

drafts/AENS.md
knarz: Allocating them for the network was an idea I had as well but we don't have a mechanism for that just yet. I'll include it, though. It would definitely be nice if the network was able to fund itself.

drafts/AENS.md
knarz: I'll add those to the document. I'd also be ok with going for ASCII only in the first version but we should definitely support the unicode mess at some point, seeing as a good amount of the people who are interested in aeternity seem to be Chinese.

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+as a suffix, i.e. `mywallet.aet`.
+
+
+### Namespaces/Labels
+
+A name is split up into labels by the `FULL STOP (U+002E .)` character.
+These labels will also be referred to as namespaces, e.g. `mywallet.aet` 
+can be understood as the `mywallet` namespace in the `aet` namespace.
+
+A label MUST be longer than three characters.
```
UlfNorell: Longer than or equal to three characters? Or are the top-level namespaces (`aet`) exempt?

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+## Open Questions:
+
+(***TODO***: Pedersen commitments, i.e. perfectly hiding, or hash commitment? 
+This also has implications for the allocation mechanism. With a simple hash
+commitment you could enable auctions on hashes, e.g. what ENS does, where people
+who are interested in a name can hold an auction without revealing the actual
+name. This would not be possible with Pedersen commitmens. On the other
+hand in the first come first serve case with hash commitments, miners could 
+keep a dictionary of names they want to snatch up whenever somebody tries to 
+register them)
```
hanssv: Would it make sense to 'salt' the namehash in the pre-commit message? Then the miners can't keep a table.

drafts/AENS.md
knarz: The `claim` transaction just finishes the claiming process, i.e. after it is in the chain, the name will be regarded as `claimed`.
So this `claim` transaction itself has no expiration date but the claimed name does.

drafts/AENS.md
knarz: That certainly would be possible but then we'd lose the ability to hold auctions on the hashes. Though, with auctions, the miners wouldn't be able to front run anymore anyway.

drafts/AENS.md
knarz: Longer than

The top-level namespaces will have some stricter rules applied to them once they become available.

drafts/AENS.md
hanssv: Maybe the salt could be public (part of the pre-commit message) we are just trying to avoid pre-computed tables in miners...

drafts/AENS.md
```diff
@@ -0,0 +1,414 @@
...
+| ttl        | 8  |
+ ------------ ----
+| pointers   |    |
+ ------------ ----
+```
+
+The `update` transaction MUST be signed by the owner
+of the name entry to be updated.
+
+The `expire_by` MUST NOT be more than 1 year into the
```
hanssv: To be consistent, maybe it is better to express this MUST NOT in terms of number of blocks ahead of current top block. Now we are mixing block height and wall clock time...

drafts/AENS.md
knarz: Hashing the labels recursively preserves the hierarchical nature of the full name, i.e.

With the given definition we go from general to specific and yours does specific to general. I think either are ok.

drafts/AENS.md
knarz: Sorry, all times should be expressed in blocks. I should make that explicit and not just assume it.

drafts/AENS.md
thepiwo: @knarz whats your take on this?

knarz: I think the simplest solution would be to forbid a `pre-claim` and the associated `claim` tx to be in the same block, or even require there to be 6 or more blocks between them. That way I can post my `pre-claim`, wait a couple of blocks and then post the claim. A malicious miner would have to mine at least two (6) blocks in a row.

I'm not sure how hashing the account together with the name would prevent a miner from just issuing their own `pre-claim` and then `claim`.

thepiwo: If the `pre-claim` is just the hash of the name, wouldn't everybody be able to pre-claim the same name if the transaction is not mined yet? When hashing the account together with the name, the name itself would be hidden but in the `claim` it can be checked if this name was actually pre-claimed by the corresponding account

drafts/AENS.md
michalzajda: why we need `.test` for testnet? the `.aet` in testnet will be really test-`.aet`?

happi: Since there will be a time (nlockheight) sometime in the future when .aet will be available we can have exactly the same code for the test net and the main net. To test nsmes before that height both domains are needed.

drafts/AENS.md
knarz: Since the account (which is just a pubkey) can be recovered from a signature, the miner could still fairly easily put their own `pre-claim` into a block before a different one.
If we want to blind the value then it has to be a nonce, which would be revealed together with the name in the `claim`. This would then mean that a user would have to rely on their client software to warn them when they try to claim an already claimed name since miners would no longer be able to tell.

Anyway, I would propose that our commitment is just `Hash(Namehash + Nonce)` for now until I figure out a good way to implement auctions. The `claim` message will then have to reveal both the hash and the nonce.

thepiwo: Maybe my thinking is of there, but the miner would not know what the preclaimed name is until revealed by the claim.

In case of the namehash he could pre-claim the same name without knowing what it is. (Now I understand the two blocks in a row by thinking about it.)

But with including the account in the namehash only the pre-claiming account could do the actual claim. A nonce would offer the same benefits.

sennui:
can we include the conclusion that aens should be integrated with existing apis?

knarz:
Sorry, what exactly do you mean?

sennui:
@knarz we recently chatted on slack, to conclude that names would be prefixed and used interchangable with hashes in our APIs (contrary to an approach, where they are resolved by the wallet and core apis would get only final hashes)

drafts/AENS.md
```diff
@@ -0,0 +1,443 @@
...
+Send a `pre-claim` transaction containing a hash commitment.
+This transaction starts the claiming process.
+
+```
+ ------------ ----
+| commitment | 32 |
+ ------------ ----
+```
+
+A `pre-claim` transaction for a name that is in the `claimed`
```
sennui: we can't check if the name is `claimed` in `pre-claim`. We will need to keep it in another MT and invalidate at `claim` level

knarz:
I'd like to keep these protocol specs free from the details of different implementations.

PR #9, opened 2018-01-12
Pt 154173179 contract spec

happi:
Inital specification of smart contracts.

### DIFF
--- a/contracts/contract_life_cycle.md
+++ b/contracts/contract_life_cycle.md
@@ -1,0 +1,3 @@
+[back](./contracts.md)
+## Contract life cycle examples
+
--- a/contracts/contract_state_tree.md
+++ b/contracts/contract_state_tree.md
@@ -1,10 +1,9 @@
 [back](./contracts.md)
 ## Contract state tree
 
-The contracts and their interactions reside in an contract state tree that
-is included in the root hash. The existence of an contract and the
-existence of query/response will have to be proven when two parites
-contract with eachother.
+The contracts and their interactions reside in a contract state tree that
+is included in the root hash. The existence of a contract and the
+existence of calls can be proven.
 
 ### Contract state tree objects
 
@@ -13,9 +12,11 @@ The contract state tree contains three types of objects:
 - The contract store
 - The contract log
 
+As well as normal account fields.
+
 #### The contract definition
 
-- Created by an contract register transaction.
+- Created by a contract create transaction.
 - Deleted when an inactive contract has no more referers.
 
 
--- a/contracts/contract_state_tree.md
+++ b/contracts/contract_state_tree.md
@@ -7,10 +7,15 @@ existence of calls can be proven.
 
 ### Contract state tree objects
 
-The contract state tree contains three types of objects:
-- The contract definition
-- The contract store
-- The contract log
+The contract state tree contains objects:
+- The contract definition (code binary)
+- The contract store (data binary)
+- The contract log (data binary)
+- Active flag (Boolean)
+- List of refering contracts
+- Calls & Return data
+- Deposit (amount)
+- vm_version
 
 As well as normal account fields.
 
--- a/contracts/contract_state_tree.md
+++ b/contracts/contract_state_tree.md
@@ -1,0 +1,26 @@
+[back](./contracts.md)
+## Contract state tree
+
+The contracts and their interactions reside in an contract state tree that
+is included in the root hash. The existence of an contract and the
+existence of query/response will have to be proven when two parites
+contract with eachother.
+
+### Contract state tree objects
+
+The contract state tree contains three types of objects:
+- The contract definition
+- The contract store
+- The contract log
+
+#### The contract definition
+
+- Created by an contract register transaction.
+- Deleted when an inactive contract has no more referers.
+
+
+### Contract state tree update
+
+
+
+### Pruning of contracts
--- a/contracts/contract_transactions.md
+++ b/contracts/contract_transactions.md
@@ -36,6 +36,8 @@ The transaction contains:
 , code            :: hex_bytes()
 , vm_version      :: hex_byte()
 , fee             :: amount()
+, deposit         :: amount()
+, amount          :: amount()
 , gas             :: amount()
 , gas_price       :: amount()
 , call_data       :: hex_bytes()
@@ -58,6 +60,15 @@ of
 ```
  "C0DE" && hash( nonce && owner)
 ```
+
+The fee, the deposit, the amount and the used gas will be
+subtracted from the owner account.
+
+The amount will be added to the contract account.
+
+The fee will added to the miners account.
+
+The deposit will be "held by the contract" until it is deactivated.
 
 The miner will add the new created contract address, the contract state
 and the return data from the initial call to the state tree.
--- a/contracts/contract_transactions.md
+++ b/contracts/contract_transactions.md
@@ -10,6 +10,12 @@ Contract transactions are of four types:
 A contract is also a normal account with a balance,
 and the normal spend transaction can be applied to the account.
 
+When a contract is createad or called, the miner checks that the
+creator/caller has gas*gas_price (*10^-18) aeons in the account.
+
+The execution of the call will use a certain amount of gas up to
+the maximum given, that amount is decduced from the caller's
+account and added to the miner's account.
 
 ### Create Contract Transaction
 
@@ -31,6 +37,7 @@ The transaction contains:
 , vm_version      :: hex_byte()
 , fee             :: amount()
 , gas             :: amount()
+, gas_price       :: amount()
 , call_data       :: hex_bytes()
 }
 ```
@@ -41,6 +48,9 @@ and hex bytes is a string of a number of hex bytes preceded with 0x.
 Call data is encoded depending on the ABI of the language of the contract.
 
 The transaction has to be signed with the private key of the owner.
+
+The special variable "caller" will be set to the same value as "owner"
+for the initial call.
 
 The miner will add the new created contract address, the contract state
 and the return data from the initial call to the state tree.
@@ -60,9 +70,13 @@ all other fields are as in create contract.
 , vm_version      :: hex_byte()
 , fee             :: amount()
 , gas             :: amount()
+, gas_price       :: amount()
 , call_data       :: hex_bytes()
 }
 ```
+
+The special variable "caller" will be set to the same value as "owner"
+for the initial call.
 
 The miner will add the return data and the state from the initial call
 to the state tree.
@@ -85,6 +99,7 @@ The transaction contains:
 , contract        :: public_key()
 , vm_version      :: hex_byte()
 , gas             :: amount()
+, gas_price       :: amount()
 , call_data       :: hex_bytes()
 , fee             :: amount()
 
--- a/contracts/contract_transactions.md
+++ b/contracts/contract_transactions.md
@@ -4,6 +4,10 @@
 Contract transactions are of three types:
 - Create
 - Call
+- Disable
+
+A contract is also a normal account with a balance.
+
 
 ### Contract create transaction
 
--- a/contracts/contract_transactions.md
+++ b/contracts/contract_transactions.md
@@ -22,11 +22,15 @@ account and added to the miner's account.
 Anyone can create a new contract by submitting a create contract transaction.
 
 The transaction contains:
-- The address of the contract owner (the one signing and paying for the transaction) 
+- The address of the contract owner (the one signing and paying for the transaction)
+- Nonce of the owner/creator account.
 - The byte code of the contract
 - The VM to use
 - A transaction fee
+- A deposit (an even number, 0 is accepted).
+- An amount (aeons to transfer to the account, 0 is accepted).
 - Gas for the initial call
+- Gas price for the call.
 - Call data for the initial call (usually including a function name and args, interpreted by the contract).
 
 
@@ -70,8 +74,13 @@ The fee will added to the miners account.
 
 The deposit will be "held by the contract" until it is deactivated.
 
+If the initial call fails (runs out of gas) the contract is not
+created.  The owner loses the fee and the gas (to the miner) but the
+amount and the deposit are return to the owner.
+
 The miner will add the new created contract address, the contract state
-and the return data from the initial call to the state tree.
+and the return data from the initial call to the state tree (if the
+init succeeds).
 
 
 ### Attach Contract Transaction
@@ -87,6 +96,8 @@ all other fields are as in create contract.
 , code            :: hex_bytes()
 , vm_version      :: hex_byte()
 , fee             :: amount()
+, deposit         :: amount()
+, amount          :: amount()
 , gas             :: amount()
 , gas_price       :: amount()
 , call_data       :: hex_bytes()
@@ -107,6 +118,8 @@ Anyone can call an existing contract (as long as it isn't disabled).
 The transaction contains:
 - The address of the caller (the one signing and paying for the transaction)
 - The address of the contract
+- An optional additional fee to the miner apart from gas.
+- Optional amount to transfer to the account before execution.
 - The amount of gas to use
 - The calldata
 - A transaction fee
@@ -116,11 +129,11 @@ The transaction contains:
 , nonce           :: pos_integer()
 , contract        :: public_key()
 , vm_version      :: hex_byte()
+, fee             :: amount()
+, amount          :: amount()
 , gas             :: amount()
 , gas_price       :: amount()
 , call_data       :: hex_bytes()
-, fee             :: amount()
-
 }
 ```
 
--- a/contracts/contract_transactions.md
+++ b/contracts/contract_transactions.md
@@ -1,5 +1,5 @@
 [back](./contracts.md)
-## Contract transactions
+## Contract Transactions
 
 Contract transactions are of three types:
 - Create
@@ -9,6 +9,60 @@ Contract transactions are of three types:
 A contract is also a normal account with a balance.
 
 
-### Contract create transaction
+### Create Contract Transaction
+
+Anyone can create a new contract by submitting a create contract transaction.
+
+The transaction contains:
+- The address of the contract owner (the one signing and paying for the transaction) 
+- The byte code of the contract
+- The VM to use
+- A transaction fee
+- Optionally: A initial call to the contract including name, args and gas. (Can be used to emulate init function of Ethereum/Solidity contracts)
+
+
+
+```
+{ owner           :: public_key()
+, code            :: hex_bytes()
+, vm_version      :: hex_byte()
+, fee             :: amount()
+, init            :: none() | { gas :: amount()
+                              , call_data :: hex_bytes()
+                              }
+}
+```
+
+Where a hex byte is a string of two hex digits preceded with 0x,
+and hex bytes is a string of a number of hex bytes preceded with 0x.
+
+Call data is encoded depending on the ABI of the language of the contract.
+
+The transaction has to be signed with the private key of the owner.
 
 ### Contract call transaction
+
+Anyone can call an existing contract (as long as it isn't disabled).
+
+The transaction contains:
+- The address of the caller (the one signing and paying for the transaction)
+- The address of the contract
+- The amount of gas to use
+- The calldata
+- A transaction fee
+
+```
+{ caller          :: public_key()
+, contract        :: public_key()
+, vm_version      :: hex_byte()
+, gas             :: amount()
+, call_data       :: hex_bytes()
+, fee             :: amount()
+
+}
+```
+
+The transaction has to be signed with the private key of the caller.
+
+The call_data is encoded in accordans with the contract language ABI.
+(See e.g. the Ring ABI.)
--- a/contracts/contract_transactions.md
+++ b/contracts/contract_transactions.md
@@ -1,0 +1,10 @@
+[back](./contracts.md)
+## Contract transactions
+
+Contract transactions are of three types:
+- Create
+- Call
+
+### Contract create transaction
+
+### Contract call transaction
--- a/contracts/contract_transactions.md
+++ b/contracts/contract_transactions.md
@@ -52,6 +52,13 @@ The transaction has to be signed with the private key of the owner.
 The special variable "caller" will be set to the same value as "owner"
 for the initial call.
 
+The contract address is created by taking the Key-length left most hex chars
+of
+
+```
+ "C0DE" && hash( nonce && owner)
+```
+
 The miner will add the new created contract address, the contract state
 and the return data from the initial call to the state tree.
 
--- a/contracts/contract_transactions.md
+++ b/contracts/contract_transactions.md
@@ -1,12 +1,14 @@
 [back](./contracts.md)
 ## Contract Transactions
 
-Contract transactions are of three types:
+Contract transactions are of four types:
 - Create
+- Attach
 - Call
 - Disable -- To be decided.
 
-A contract is also a normal account with a balance.
+A contract is also a normal account with a balance,
+and the normal spend transaction can be applied to the account.
 
 
 ### Create Contract Transaction
@@ -18,18 +20,18 @@ The transaction contains:
 - The byte code of the contract
 - The VM to use
 - A transaction fee
-- Optionally: A initial call to the contract including name, args and gas. (Can be used to emulate init function of Ethereum/Solidity contracts)
-
+- Gas for the initial call
+- Call data for the initial call (usually including a function name and args, interpreted by the contract).
 
 
 ```
 { owner           :: public_key()
+, nonce           :: pos_integer()
 , code            :: hex_bytes()
 , vm_version      :: hex_byte()
 , fee             :: amount()
-, init            :: none() | { gas :: amount()
-                              , call_data :: hex_bytes()
-                              }
+, gas             :: amount()
+, call_data       :: hex_bytes()
 }
 ```
 
@@ -39,6 +41,32 @@ and hex bytes is a string of a number of hex bytes preceded with 0x.
 Call data is encoded depending on the ABI of the language of the contract.
 
 The transaction has to be signed with the private key of the owner.
+
+The miner will add the new created contract address, the contract state
+and the return data from the initial call to the state tree.
+
+
+### Attach Contract Transaction
+
+Attach contract code to an existing accont.
+In this case owner == the acount addrres and no new account will be created,
+all other fields are as in create contract.
+
+
+```
+{ owner           :: public_key()
+, nonce           :: pos_integer()
+, code            :: hex_bytes()
+, vm_version      :: hex_byte()
+, fee             :: amount()
+, gas             :: amount()
+, call_data       :: hex_bytes()
+}
+```
+
+The miner will add the return data and the state from the initial call
+to the state tree.
+
 
 ### Contract call transaction
 
@@ -53,6 +81,7 @@ The transaction contains:
 
 ```
 { caller          :: public_key()
+, nonce           :: pos_integer()
 , contract        :: public_key()
 , vm_version      :: hex_byte()
 , gas             :: amount()
@@ -66,3 +95,6 @@ The transaction has to be signed with the private key of the caller.
 
 The call_data is encoded in accordance with the contract language ABI.
 (See e.g. the Ring ABI.)
+
+The miner will add the return data and the state of the call to the state
+tree.
--- a/contracts/contract_transactions.md
+++ b/contracts/contract_transactions.md
@@ -64,5 +64,5 @@ The transaction contains:
 
 The transaction has to be signed with the private key of the caller.
 
-The call_data is encoded in accordans with the contract language ABI.
+The call_data is encoded in accordance with the contract language ABI.
 (See e.g. the Ring ABI.)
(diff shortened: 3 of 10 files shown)


contracts/contract_state_tree.md
```diff
@@ -0,0 +1,26 @@
+[back](./contracts.md)
+## Contract state tree
+
+The contracts and their interactions reside in an contract state tree that
+is included in the root hash. The existence of an contract and the
+existence of query/response will have to be proven when two parites
```
gorillainduction: parites -> parties

gorillainduction: query/response?

contracts/contract_transactions.md
```diff
@@ -0,0 +1,68 @@
...
+- Optionally: A initial call to the contract including name, args and gas. (Can be used to emulate init function of Ethereum/Solidity contracts)
+
+
+
+```
+{ owner           :: public_key()
+, code            :: hex_bytes()
+, vm_version      :: hex_byte()
+, fee             :: amount()
+, init            :: none() | { gas :: amount()
```
lucafavatella: Instead of `none()` you mean `undefined`? I do not see how value can have type `none()`.

happi: Perhaps.
Well it is something like option(X) or perhaps use [] for optional?

We do not really have defined language for our transaction types or do we?

Suggestions are welcome.

happi: I really mean that the whole init field could be missing.

contracts/contracts.md
```diff
@@ -0,0 +1,54 @@
...
+- A contract runs on a virtual machine.
+- A contract is owned by an **contract owner**.
+- The contract owner creates a contract through posting a **contract creation transaction** on the chain.
+- The contract creation transaction register an account as a contract. (One account - one contract)
+- Any user can call an exported function in a contract by posting a  **contract call transaction** on the chain.
+- Contract can be written in a high level language which is compiled to the VM bytecode.
+
+## [Contract life cycle examples](./contract_life_cycle.md)
+
+There is no kill instruction in the Aetherium virutal machine, instead there is a deactivation instruction.
```
lucafavatella: Typos?

contracts/contract_transactions.md
lucafavatella: > I really mean that the whole init field could be missing.

ah ok

contracts/contracts.md
happi: Thanks. Going over things with a spellchecker now.

contracts/contract_transactions.md
```diff
@@ -0,0 +1,68 @@
...
+
+Contract transactions are of three types:
+- Create
+- Call
+- Disable
+
+A contract is also a normal account with a balance.
+
+
+### Create Contract Transaction
```
gorillainduction: What is the address of the contract? Is it set by the owner in the create transaction?

gorillainduction: Can the account of the contract act by itself (e.g., a spend transaction)? If not, we must check if something is a contract on all other transactions.

If the account of a contract can make other transactions, can it for example register itself as an oracle?

Or is it only the owner that can make a spend transaction from the account? Then this is a delegated signing, i.e., someone else signs the transaction for an account. We don't have that currently.

contracts/contract_transactions.md
```diff
@@ -0,0 +1,68 @@
...
+
+Anyone can create a new contract by submitting a create contract transaction.
+
+The transaction contains:
+- The address of the contract owner (the one signing and paying for the transaction) 
+- The byte code of the contract
+- The VM to use
+- A transaction fee
+- Optionally: A initial call to the contract including name, args and gas. (Can be used to emulate init function of Ethereum/Solidity contracts)
```
gorillainduction: If the init fails, is the contract created?

contracts/contract_transactions.md
```diff
@@ -0,0 +1,68 @@
...
+
+Call data is encoded depending on the ABI of the language of the contract.
+
+The transaction has to be signed with the private key of the owner.
+
+### Contract call transaction
+
+Anyone can call an existing contract (as long as it isn't disabled).
+
+The transaction contains:
```
gorillainduction: Should it not also have a transfer of coin?

contracts/contracts.md
```diff
@@ -0,0 +1,54 @@
+# Smart Contracts
+
+## Overview and definitions
+
+- A smart **Contract** is a program on the blockchain that lives in the **contract state tree** in a full node.
+- A contract runs on a virtual machine.
+- A contract is owned by an **contract owner**.
+- The contract owner creates a contract through posting a **contract creation transaction** on the chain.
```
gorillainduction: Should it be 'contract create transaction'?

contracts/contracts.md
```diff
@@ -0,0 +1,54 @@
...
+When a contract is disabled it can only be called from other contracts already created on the chain.
+One can not create a new contract on the chain that calls a disabled contract neither directly nor indirectly.
+
+A typically contract life cycle looks like this:
+1. Contract A with a function f is created by owner O.
+2. Anyone can call the function f in A (call A.f()) any number of times.
+3. Another contract B that refers to A.f is created.
+4. Contract A is deactivated by O.
+5. No one can call A directly any more, and no new contract can refer to contracs A and B, but calls though B still goes to A.f.
+6. Contract B is deactivated by its owner, and has no refering contracts, it is dealocated as is contract A. (removed form the contract state tree)
```
gorillainduction: dealocated -> deallocated

contracts/contract_transactions.md
gorillainduction: For oracles, an account is first created (by the owner), and then it registers itself as an oracle. This might be a good solution for contracts at the first step.

This can later be extended with a dual signing by the contract (to ensure it can operate) and the creator.

This also removes the need for an owner of the contract. The contract itself will sign it's disabling.

contracts/contract_transactions.md
gorillainduction: Don't use the none() type anyway, since it has a specific meaning. We can have a discussion whether it is good to have missing (optional) fields in a transaction at some point as well.

contracts/contract_life_cycle.md
```diff
@@ -0,0 +1,3 @@
+[back](./contracts.md)
+## Contract life cycle examples
```
lucafavatella: Shall this file be suppressed? I see "contracts/contracts.md" links to this doc though the lifecycle example is there.

contracts/contract_transactions.md
```diff
@@ -0,0 +1,68 @@
...
+
+A contract is also a normal account with a balance.
+
+
+### Create Contract Transaction
+
+Anyone can create a new contract by submitting a create contract transaction.
+
+The transaction contains:
+- The address of the contract owner (the one signing and paying for the transaction)
```
lucafavatella: plus nonce I guess

contracts/contract_transactions.md
happi: Thinking a bit more about it it isn't clear whether it should be allowed to not run any code on contract creation.

contracts/contract_transactions.md
happi: The address of the contract is created by the create transaction it is a new unique address. (At least that was my thought.)

contracts/contract_life_cycle.md
happi: The plan is to add more examples.

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+
+A Ring contract has a state. You must declare the state record type in
+the contract. You can then update the state through mutable record
+update.
+
+```
+type state = { mutable field: string
+             , mutable value: uint};
+let state = {field: “”, value : 0};
+let store_value = (f: string, v: uint) => {
```
rjmh: This isn't compatible with the statement above that state is updated by calling put.

contracts/contracts.md
```diff
@@ -0,0 +1,54 @@
...
+There is no kill instruction in the Aeternity virtual machine, instead there is a deactivation instruction.
+When a contract is disabled it can only be called from other contracts already created on the chain.
+One can not create a new contract on the chain that calls a disabled contract neither directly nor indirectly.
+
+A typically contract life cycle looks like this:
+1. Contract A with a function f is created by owner O.
+2. Anyone can call the function f in A (call A.f()) any number of times.
+3. Another contract B that refers to A.f is created.
+4. Contract A is deactivated by O.
+5. No one can call A directly any more, and no new contract can refer to contracts A and B, but calls though B still goes to A.f.
```
lucafavatella: though -> through ?
goes -> go ?

contracts/contracts.md
```diff
@@ -0,0 +1,54 @@
...
+6. Contract B is deactivated by its owner, and has no referring contracts, it is deallocated as is contract A. (removed form the contract state tree)
```
lucafavatella: When contract A was created, it had no referring contracts. How is it A was not deallocated as soon as it was created?

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+| Type    | Description                     | Example
+| ------- | ------------------------------- | -------:
+| uint    | A 256 bit integer               | ```42```
+| int     | A 256 bit 2-complement integer  | ```-1```
+| address | A 256 bit number given as a hex | ```ff00```
+| bool    | A Boolean                       | ```true```
+| string  | An array of bytes               | ```“Foo”```
+| list    | A homogeneous immutable singly linked list. | ```[1, 2, 3]```
+| tuple   | An ordered heterogeneous array   | ```(42, “Foo”, true)```
+| record  | An immutable key value store with fixed key names and typed values | ``` type balance = { owner: address, value: uint } ```
```
lucafavatella: Is order of keys in definition relevant?

contracts/contracts.md
lucafavatella: I see - it needs to be disabled explicitly.

contracts/contract_transactions.md
gorillainduction: This needs more discussion. We should probably do it offline.

contracts/ring.md
happi: No. But in the implementation/internal representation it has to be defined, and I think sorted on key is the way to go.

contracts/contract_transactions.md
```diff
@@ -0,0 +1,68 @@
...
+- Disable -- To be decided.
+
+A contract is also a normal account with a balance.
+
+
+### Create Contract Transaction
+
+Anyone can create a new contract by submitting a create contract transaction.
+
+The transaction contains:
```
lucafavatella: Shall this contain also a (even) deposit? See contracts/ring.md "killing a contract"

contracts/contract_transactions.md
```diff
@@ -0,0 +1,68 @@
...
+Call data is encoded depending on the ABI of the language of the contract.
+
+The transaction has to be signed with the private key of the owner.
+
+### Contract call transaction
+
+Anyone can call an existing contract (as long as it isn't disabled).
+
+The transaction contains:
+- The address of the caller (the one signing and paying for the transaction)
```
lucafavatella: plus nonce I guess

contracts/contract_transactions.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+Where a hex byte is a string of two hex digits preceded with 0x,
+and hex bytes is a string of a number of hex bytes preceded with 0x.
+
+Call data is encoded depending on the ABI of the language of the contract.
+
+The transaction has to be signed with the private key of the owner.
+
+### Contract call transaction
```
lucafavatella: Notes on handling of failures may be beneficial.

Cases of failures:
* gas in tx turns out not to be enough
* referred contract is disabled (maybe it was enabled when caller signed this tx)
* referred contract does not exist (maybe it was existed when caller signed this tx)
* call aborts because of bad bytecode
* call aborts/fails because some of the assumptions in the code fail (e.g. purely speculatively for examples - a referred oracle does not exist, a referred oracle has unexpected type, an account balance is not high enough for debiting)

happi: How are failures for spend transactions handled?
Are they signaled some how or just left in the tx-pool?

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+
+The reference count of a contract is handled as the account balance
+and kept in the state tree of the miner and the merkle hash is
+included in the state hash in each block just as with balances.
+
+The transaction for creating a contract has an extra fee called
+deposit which has to be an even number. The disable transaction is
+free but the miner and the creator get half of the deposit fee each at
+contract disable thus encouraging creators to disable their contracts
+and miners to pick disable transactions.
```
knarz: Wouldn't a miner have an inherent incentive to pick up disable transactions since that means they're going to have to potentially store/process less data?
And would the size of this deposit be mandated by the network or freely chosen?

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+```
+balance() : () => uint
+```
+The amount of coins currently in the contract account
+```
+contract_address() : () => address;
+```
+The address of the contract.
+
+Block references
```
knarz: Is it possible the access previous blocks?

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+    state.value = v;
+}
+```
+#### Contract primitives
+```
+creator() : () => address
+```
+The address (pubkey) of the entity signing for the contract creation transaction
+```
+caller() : () => address
```
knarz: If I call contract A, which then calls contract B, what would `caller()` return in contract B? The address of A, I assume?

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+- The value of the state is accessible from inside the contract
+  through an implicitly bound variable state.
+- State updates are performed by calling a function put : state =>
+  unit (possibly with a better name).
+- Aside from the put function (and similar functions for transactions
+  and events), the contract language is purely functional.
+- Functions modifying the state need to be annotated with the stateful keyword.
+
+To make it convenient to update parts of a deeply nested state Ring
+provides special special syntax for map/record updates.  Open
```
knarz: -special

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
+[back](./contracts.md)
+## The Ring Language
+An AEternity BlockChain Language
+The Ring is a dialect of ReasonML ( https://reasonml.github.io/ ).
+
+Ring is customized for smart contracts, that has the potential of
+being published to a blockchain (the AEternity BlockChain). Thus some
+unnecessary features of Reason (and OCaml - since Reason is closely
+related to OCaml) has been removed, and some blockchain specific
```
knarz: s/has/have/

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
+[back](./contracts.md)
+## The Ring Language
+An AEternity BlockChain Language
+The Ring is a dialect of ReasonML ( https://reasonml.github.io/ ).
+
+Ring is customized for smart contracts, that has the potential of
```
knarz: s/that has the potential of being/which can potentially be/

contracts/contract_transactions.md
```diff
@@ -0,0 +1,68 @@
...
+
+Anyone can call an existing contract (as long as it isn't disabled).
+
+The transaction contains:
+- The address of the caller (the one signing and paying for the transaction)
+- The address of the contract
+- The amount of gas to use
```
knarz: Is this going to have the same semantics are ethereum? i.e. I specify the gas price and the gas? Will not used gas be refunded as well?

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+contracts. Contracts are similar to classes in object oriented
+languages and support local state and inheritance. More specifically:
+
+- A contract implementation, or simply a contract, is the code for a
+  smart contract and consists of a list of types and functions that
+  may or may not have a definition. A contract where all types and
+  functions are defined is called a concrete contract. Only concrete
+  contracts can be instantiated. Previously we had a notion of
+  contract type, which is now replaced by fully abstract contracts
+  where no types or functions have definitions. Example:
```
UlfNorell: Drop the "Previously ...". Maybe something like

> A fully abstract contract, where no types of functions have definitions, can be used to represent a contract type. Example:

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+| list    | A homogeneous immutable singly linked list. | ```[1, 2, 3]```
+| tuple   | An ordered heterogeneous array   | ```(42, “Foo”, true)```
+| record  | An immutable key value store with fixed key names and typed values | ``` type balance = { owner: address, value: uint } ```
+| map     | An immutable key value store with dynamic mapping of keys of one type to values of one type | ```type accounts = map(string, address)```
+| state   | A record of blockstate key, value pairs  |
+| transactions | An append only list of blockchain transactions |
+| events   | An append only list of blockchain events (or log entries) |
+
+#### Arrays
+Ring might not have arrays… Ring might have Maps.
```
UlfNorell: Lose this?

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+
+####  Variants
+Depending on how we write the compiler, if they are cheap to include
+some like to use variant types, while others might not want them.
```
UlfNorell: In a statically typed language you absolutely need variants.

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+(parameters). Keep separate from mutable state or annotate certain
+fields as immutable?
+
+### Types
+Ring has the following types:
+
+| Type    | Description                     | Example
+| ------- | ------------------------------- | -------:
+| uint    | A 256 bit integer               | ```42```
+| int     | A 256 bit 2-complement integer  | ```-1```
```
UlfNorell: Is there a good reason to have 256-bit integers rather than arbitrary precision integers?

UlfNorell: Do we absolutely need both signed and unsigned integers? Overloading arithmetic operations will complicate type checking and compilation.

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+we definitely want it in the final language.
+
+### Builtins
+#### Events
+Ring has events which logs a structured message to the contract log in
+the resulting blockchain transaction.
+
+```
+event(name: string, value: any)
+```
```
UlfNorell: We probably want event types to be declared.

The simplest option would be that a contract can declare a single `event` type (which can be a variant type) and the event function takes an argument of this type.

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+
+#### Transactions
+Ring can generate blockchain transactions.
+```
+transaction(tx_type: transaction, argument: record )
+```
```
UlfNorell: Todo: specify the transaction type. The arguments should probably be part of the transaction (different transaction types have different arguments).

Question: should you be able to inspect transactions?

contracts/ring.md
UlfNorell: Indeed. This section is left-over from a previous version of the spec.

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+type state = { mutable field: string
+             , mutable value: uint};
+let state = {field: “”, value : 0};
+let store_value = (f: string, v: uint) => {
+    state.field = f;
+    state.value = v;
+}
+```
+#### Contract primitives
+```
```
UlfNorell: Rather than having separate functions for each primitive, I think we should have a few (three?) records containing the necessary information. Something like

```
call : {caller : address, value : uint, ...}
contract : {creator : address, balance : uint, address : address, ...}
chain : {height : uint, timestamp : uint, ...}
```

contracts/ring.md
```diff
@@ -0,0 +1,340 @@
...
+PUSH   Arity + 1 ; # words to skip and
+DUP              ; Size = calldatasize - (arity+1)
+CALLDATASIZE  
+SWAP1
+SUB         
+CALLDATACOPY     ; copy from [arity+1..Size] to mem(0)
+
+JUMPDEST         ; for local calls
+                 ; code for function body
+```
```
UlfNorell: Wouldn't it be simpler (and safer) to do the entry point computation outside the byte code? Each contract would store a list of entry points, and when calling a contract we look up the entry point for the called function and start execution from there. That would save a lot of boiler plate in the contract byte code.

Of course, there's the question of Solidity compatibility.

contracts/contract_transactions.md
happi: Ok so create will execute the contract and the create transaction will take a calldata field which for ring should make the contract run an init function.

contracts/contract_transactions.md
happi: Update after design meeting:
A contract and an account is the same thing.
The create contract instruction will create an account with a new pubkey that no one has the private key to.
There will be an attach contract transaction that will attach code to a "normal" account. This is so that a user can e.g. implement his own wallet functionality.
I will update the spec.

contracts/contract_transactions.md
happi: If it runs out of gas or similar I guess the contract creation should be considered to fail.

contracts/contract_transactions.md
happi: Yes.

contracts/contract_transactions.md
happi: Yes. Fixed.

contracts/contract_state_tree.md
```diff
@@ -0,0 +1,27 @@
...
+The contracts and their interactions reside in a contract state tree that
+is included in the root hash. The existence of a contract and the
+existence of calls can be proven.
+
+### Contract state tree objects
+
+The contract state tree contains three types of objects:
+- The contract definition
+- The contract store
+- The contract log
```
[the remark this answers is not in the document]
happi: Yes. I'll add.

contracts/ring.md
happi: The word size in the current VM  (EVM) is 256-bit. Implementing arbitrary precision on that VM will be tricky/costly.
Since the argument arrive as just binaries I think knowing whether it is signed or unsigned will help.

contracts/ring.md
happi: @UlfNorell Do you have a new version?

contracts/contract_transactions.md
lucafavatella: > How are failures for spend transactions handled?
> Are they signaled some how or just left in the tx-pool?

[Current handling](https://www.pivotaltracker.com/story/show/154323788/comments/184659656) of those txs is that they are not included in the block and are left in the mempool.

zp-sd: Regarding transaction failures, I think an idea was to keep such transactions in the pool, as failed transactions may get valid when a fork takes over (e.g. on a fork, an account may have sufficient funds to cover spend).

But this was intended to be coupled with some mempool pruning based on defined TTL of all transactions (how long they may live in the mempool) and max limit of transactions kept in the mempool.

_(all this is just what I remember from the initial design, not sure if this has changed)_

contracts/ring.md
happi: Yes, I have been thinking about this.
Solidity could call address 0 and would work as before.
There are only two questions:
1. How do we call a function in another contract? We would have to extend the call function in the VM with an address.
2. Who and when is the call address calculated? How does the caller get to know the call address, which is dependent on the bytecode. (in practice that is)

UlfNorell: If we want to keep Solidity compatibility, I guess we'd add a new call instruction with the extra argument. If we pass it the index of the entry point, rather than the address, we'd be less dependent on the actual byte code.

contracts/ring.md
UlfNorell: The new version is already included at [the top](https://github.com/aeternity/protocol/blob/fa0c94983f6f4550d2de55ea7fa0c38ee2ab3507/contracts/ring.md#mutable-state).

contracts/ring.md
UlfNorell: The inherent incentive would be very weak though. If it's more immediately profitable to pick other transactions, a miner wouldn't pick disable transactions until they're starting to get low on memory/disk.

I'm more concerned about malicious disabling of contracts. If disabling a contract is a separate transaction that seems difficult to prevent. Wouldn't it make more sense to only allow the contract code to disable itself? At least that way a user of a contract has the ability to check under which conditions the contract can be disabled.

contracts/ring.md
happi: Hmm.
"State updates are performed by calling a function put : state =>
  unit (possibly with a better name)."
Does the function put only take a new state? Or does it take a field name and a value also?
If it only takes a state how do you create the new state syntactically? (as in the above description? how do you just update one of the fields?)
Some examples would be nice.

contracts/contract_transactions.md
happi: Yes. Fixed.

contracts/contract_transactions.md
happi: Yes.

contracts/ring.md
happi: Yes, this is a valid point (and the reason the disable transaction is listed as "to be decided").

@knarz What is your view on having a disable transaction or only a disable instruction in the VM?

contracts/ring.md
happi: Yes, as far as the VM allows. I am not sure how to express that in the langugage.
What does Solidity do?

contracts/ring.md
happi: Looks fine to me.

contracts/ring.md
UlfNorell: There are some syntax example in the repo, but we should definitely describe it more precisely (and refine it). Briefly, we have syntax to create and update records:
```
  let newRec = {x: 42, y: "bar"};
  let updRec = newRec {x: 43};
```
It also allows deep updates
```
  let r2 = {r: updRec};
  let r2' = r2 {r.x: 44};
```
The parser doesn't yet handle it, but I want to use the same notation for maps, so you can write `m { [key]: val }` to set the value of `key` to `val` in the map `m`.

Also I think I prefer `=` over `:`, which is now possible since we don't have assignment.

So, to answer your question, you would write something like
```
  put(state { stakeholders[caller].balance: b })
```
to update a part of the state. There's some hope that the compiler would be able to compile this to an inplace memory write.

contracts/ring.md
happi: Yes, and if we limit the number of functions to 256 we only need one byte. Function 256 could switch on an extra arguments if needed. OK?

UlfNorell: Sounds good to me.

contracts/ring.md
knarz: Solidity provides a special, global variable `blocks`, which allows to get the hashes of the previous 256 blocks and also some properties of the current block, [see here](https://solidity.readthedocs.io/en/develop/units-and-global-variables.html#block-and-transaction-properties)

contracts/ring.md
keypair: @happi why is this the case? what do we do better then the evm?

epoch/api/contract_api_usage.md
```diff
@@ -0,0 +1,45 @@
...
+* HTTP external API: 3013
+* HTTP internal API: 3113
+
+There are two basic types of API calls, off-chain operations and on-chain transactions.
+
+In order to affect the state of the chain you have to submit a signed transaction
+to a mining node.
+
+For an up to date list of all HTTP API endpoints and their arguments see
+[Epoch HTTP API](https://aeternity.github.io/epoch-api-docs/?config=https://raw.githubusercontent.com/aeternity/epoch/master/apps/aehttp/priv/swagger.json)
```
lucafavatella: I suggest removing this paragraph as it links to master - so breaks tagging - and anyway list of APIs shall be linked from release notes.

epoch/api/contract_api_usage.md
```diff
@@ -0,0 +1,45 @@
+[back](./README.md)
+# Contracts - intended usage
```
lucafavatella: It would be nice mentioning in this intended usage doc a sample contract - or versioned link to working sample contract.

epoch/api/contract_api_usage.md
```diff
@@ -0,0 +1,45 @@
...
+For an up to date list of all HTTP API endpoints and their arguments see
+[Epoch HTTP API](https://aeternity.github.io/epoch-api-docs/?config=https://raw.githubusercontent.com/aeternity/epoch/master/apps/aehttp/priv/swagger.json)
+
+## Off chain operations
+
+An epoch node provides some utility functions to help you create contract transactions.
+
+
+### Contract Compile 
+[/contract/compile doc](https://aeternity.github.io/epoch-api-docs/?config=https://raw.githubusercontent.com/aeternity/epoch/master/apps/aehttp/priv/swagger.json#/external/CompileContract)
```
lucafavatella: I am a bit concerned re linking master from this repo, as this would not play nicely with tagging this repo. Though as a draft may be ok.

happi: Or some other way to refer to the generated doc?

epoch/api/contract_api_usage.md
happi: I think having links from protocol/epoch API descriptions to the generated specific API documentation is a good thing, but there should perhaps be a nice way to link to the documentation of the latest release.

epoch/api/contract_api_usage.md
lucafavatella: Will polish afterwards. Thanks.

PR #11, opened 2018-01-17
Fix links in epoch_api/oracle_api.md

hanssv:


lucafavatella:
I see two approvals, merging.

PR #12, opened 2018-01-17
Ensure separation of doc re API specific to epoch node

lucafavatella:


hanssv:
Looks good to me, you have a smaller textwidth that might lead to future reformatting but not important now...

PR #21, opened 2018-01-23
Chain WS API

velzevur:
Chain WS actions:
* get block by height/hash

* get header by height/hash

* push notification for mining a block

### DIFF
--- a/epoch/api/README.md
+++ b/epoch/api/README.md
@@ -28,7 +28,26 @@ The epoch node uses the plain HTTP protocol for its API endpoints during the tes
 
 ## WebSocket API definition
 
+### General Request/Response types
+A request has the format:
+
+| Name | Type | Description | Required |
+| ---- | ---- | ----------- | -------- |
+| target | string | what component to target | Yes |
+| action | string | what is the action | Yes |
+| payload | object | data for action | Yes |
+
+A response (and an event) has the format:
+
+| Name | Type | Description | Required |
+| ---- | ---- | ----------- | -------- |
+| origin | string | what component did the action | Yes |
+| action | string | what was the action | Yes |
+| payload | object | data from the action | Yes |
+
+### List of WS APIs
 * [Oracle WS API](./oracle_ws_api.md)
+* [Chain WS API](./chain_ws_api.md)
 
 ## User API - intended usage
 
--- a/epoch/api/README.md
+++ b/epoch/api/README.md
@@ -28,6 +28,36 @@ The epoch node uses the plain HTTP protocol for its API endpoints during the tes
 
 ## WebSocket API definition
 
+### Connection
+The epoch node supports an endpoint with a configurable port where the
+websocket's clients connect. It is located on `/websocket`.
+
+The node could serve multiple websocket clients. Their number is configured in
+the epoch.yaml. When all websocket connections are consumed - any new incoming
+connections will be queued. The queue has a maximum size and when it is
+reached - any new incoming connections will be rejected with an error code 400.
+This is to prevent the node of being overloaded with websocket connections.
+
+### Subscription to events
+Websocket clients will receive updates to events they've subscribed to. A
+client can subscribe to certain events on connecting and to others - on the
+fly.
+
+An example for the former can be [mining a block](./chain_ws_api.md#block-mined) event. On connection a client can pass a list of services to pass with a query parameter `subscribe_to`. If it is not passed - the client is subscribed to all events:
+
+* [mining a block](./chain_ws_api.md#block-mined) - keyword: `block_created`
+
+All events to be subscribed to are to be comma separated. Keyword `all` has
+the same effect as not passing the parameter at all. Examples:
+```
+/websocket?subscribe_to=all
+/websocket?subscribe_to=block_created
+/websocket?subscribe_to=event1,event2
+```
+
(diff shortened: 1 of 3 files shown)


velzevur:
This is the current state of WS chain interactions. Will discuss with Emil if they want the tx_objects added to the WS actions too (as we did for the HTTP APIs)

epoch/api/README.md
```diff
@@ -28,7 +28,26 @@ The epoch node uses the plain HTTP protocol for its API endpoints during the tes
...
+| target | string | what component to target | Yes |
+| action | string | what is the action | Yes |
+| payload | object | data for action | Yes |
+
+A response (and an event) has the format:
+
+| Name | Type | Description | Required |
+| ---- | ---- | ----------- | -------- |
+| origin | string | what component did the action | Yes |
+| action | string | what was the action | Yes |
```
lucafavatella: I understand this PR documents status quo of WS API, so the following comment is not meant to be addressed in this PR.
This looks like a request-response protocol on top of WS. I wonder what the programming model is: can client send second request before response to first request arrive? If yes, can responses arrive out of order? If yes, do we then need random ref in request for correlating response?

epoch/api/chain_ws_api.md
```diff
@@ -0,0 +1,253 @@
...
+  | height | number | Height of the header to fetch | yes |
+
+### Response
+ * **origin:** `chain`
+ * **action:** `requested_data`
+ * **payload:**
+
+  | Name | Type | Description | Required |
+  | ---- | ---- | ----------- | -------- |
+  | block | object - [header](#header) |  | Yes |
```
lucafavatella: Dead link?

epoch/api/README.md
```diff
@@ -28,7 +28,26 @@ The epoch node uses the plain HTTP protocol for its API endpoints during the tes
 
 ## WebSocket API definition
 
+### General Request/Response types
```
lucafavatella: Does paragraph need to be added on presence of "responses without requests" (or however we call them)? And shall the subscription/unsubscription logic be described in each of them? (I am trying to understand how subscription to documented response `mined_block` comes from.)

lucafavatella: > And shall the subscription/unsubscription logic be described in each of them?

I correct myself: as the the separate list of files e.g. `chain_ws_api.md` are meant to be pseudo-schemas (rather than intended usage), maybe that is not the correct place. Though this may need to be documented somewhere.

epoch/api/chain_ws_api.md
```diff
@@ -0,0 +1,253 @@
+[back](./README.md)
+# Chain WebSocket API
+
+The WebSocket API provides the following actions:
+* [Get a block by height](#get-block-by-height)
+* [Get a block by hash](#get-block-by-hash)
+* [Get a header by height](#get-a-header-by-height)
```
lucafavatella: Deal link?

epoch/api/chain_ws_api.md
```diff
@@ -0,0 +1,253 @@
...
+* [Get a header by hash](#get-a-header-by-hash)
```
lucafavatella: Dead link?

epoch/api/README.md
velzevur: A client can send as many requests as one likes and responses will arrive in any order. A desicion had been made that no state changing requests should be made via WS and all requests should be for immutable data (for example getting a block).
We shouldn't be having requestIDs in websockets

epoch/api/README.md
velzevur: Currently every single websocket client is subscribed to the `mine_block` events

lucafavatella: Maybe adding a note in ["User API - intended usage"](https://github.com/aeternity/protocol/pull/21/files#diff-8b16649bc0188ba5b35d5f92f6b8656aR52) listing all of the notifications the WS client is implicitly subscribed to would help the reader.

epoch/api/README.md
```diff
@@ -28,6 +28,36 @@ The epoch node uses the plain HTTP protocol for its API endpoints during the tes
...
+
+All events to be subscribed to are to be comma separated. Keyword `all` has
+the same effect as not passing the parameter at all. Examples:
+```
+/websocket?subscribe_to=all
+/websocket?subscribe_to=block_created
+/websocket?subscribe_to=event1,event2
+```
+
+Example for specific events to be subscribed on the fly are the Orcale ones as
```
lucafavatella: orcale typo

epoch/api/README.md
```diff
@@ -28,7 +28,56 @@ The epoch node uses the plain HTTP protocol for its API endpoints during the tes
...
+websocket's clients connect. It is located on `/websocket`.
+
+The node could serve multiple websocket clients. Their number is configured in
+the epoch.yaml. When all websocket connections are consumed - any new incoming
+connections will be queued. The queue has a maximum size and when it is
+reached - any new incoming connections will be rejected with an error code 400.
+This is to prevent the node of being overloaded with websocket connections.
+
+### Subscription to events
+Websocket clients will receive an updates to events they've subscribed. A
```
hanssv: Bad sentence. Remove 'an' and add 'to' after subscribed.

velzevur:
Fixed a typo and modified `epoch/api/chain_ws_api.md` according to @hanssv advice

PR #48, opened 2018-02-08
Update document to reflect change from SHA256 -> Blake2b256

hanssv:


knarz:
That should probably be `Blake2b`, since the `s` and `b` flavours exist. Also, `Blake2b` default output is also 512 bits, whereas we use 256 bits, which should also be specified.

hanssv:
Blake2b everywhere and a couple of mentions of 256 bits digest

PR #49, opened 2018-02-09
Fix also the examples not just the spec for Oracle subscriptions

hanssv:


lucafavatella:
I see this is still applicable to epoch master. I see two approvals. Merging.

PR #56, opened 2018-03-21
New serialization format of objects

gorillainduction:


serializations.md
```diff
@@ -0,0 +1,321 @@
...
+| Fieldname | Size (bytes) |
+| --- | --- |
+| height       | 8    |
+| prev_hash    | 32   |
+| txs_hash     | 32   |
+| root_hash    | 32   |
+| target       | 8    |
+| nonce        | 8    |
+| time         | 8    |
+| version      | 8    |
```
happi: Wouldn't it make sense to have version as the first field to make it possible to change the field after version in future versions?

happi: That is if we want to be able to reuse the hashing serialization for sending header and blocks to other nodes, which I think would be a good idea.

gorillainduction: That would make sense. This part is only copied from the current documentation, but we could change this as well.

serializations.md
```diff
@@ -0,0 +1,321 @@
+# Serialization formats
+
+The serialization formats in this document describes the binary format
+used to encode the different objects in the binary format that is used
+for hashing. Other formats may be used for communication between
+nodes.
```
lucafavatella: .... or for the user API.

serializations.md
```diff
@@ -0,0 +1,321 @@
...
+```
+[ <tag>, <version>, <field1>, <field2>...]
+```
+
+Since all values are byte arrays in RLP, `int()` needs to be a byte
+array as well. We encode all integers as unsigned, big endian byte
+arrays. To avoid ambiguity in the encoding of integers, we adapt the
+same scheme as RLP and demand that ingegers are encoded with the
+minimal number of bytes (i.e., disallow leading zeroes in the encoded
+byte array).
```
lucafavatella: Are negative integer numbers representable? If yes, are +0 and -0 two distinct representations?

serializations.md
```diff
@@ -0,0 +1,321 @@
...
+| nonce        | 8    |
+| time         | 8    |
+| version      | 8    |
+| pow_evidence | 168  |
+
+
+#### Block
+
+The only difference between a block and its header is the list of
+transactions. The transactions are captured in the header by the
```
lucafavatella: ... and - in a sense - the state trees.

serializations.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+Since all values are byte arrays in RLP, `int()` needs to be a byte
+array as well. We encode all integers as unsigned, big endian byte
+arrays. To avoid ambiguity in the encoding of integers, we adapt the
+same scheme as RLP and demand that ingegers are encoded with the
+minimal number of bytes (i.e., disallow leading zeroes in the encoded
+byte array).
+
+### Binary serialization
```
lucafavatella: I understand this is for:
* Objects as in state trees;
* Txs as hashed for txs tree;
* Txs as signed.

serializations.md
lucafavatella: Version is already the first in the code https://github.com/aeternity/epoch/blob/v0.9.0/apps/aecore/src/aec_headers.erl#L109

serializations.md
uwiger: @lucafavatella, as I understand RLP encoding, negative numbers are out of scope entirely.

From the [RLP Wiki](https://github.com/ethereum/wiki/wiki/RLP):
> The only purpose of RLP is to encode structure; encoding specific data types (eg. strings, floats) is left up to higher-order protocols; but positive RLP integers must be represented in big endian binary form with no leading zeroes (thus making the integer value zero be equivalent to the empty byte array).

gorillainduction: Integers are only present in RLP as size fields. You can't actually encode integers as values in RLP.

As we don't need negative integers in the system, I chose to have them unsigned. I think this makes most sense for compactness. If we do need negative integers at some point, this is easily extendable.

serializations.md
gorillainduction: The state tree is represented in the header as well, and the state tree is as connected to the header as to the block. I think this reads well as is.

serializations.md
gorillainduction: I added this at the top of the document.

PR #63, opened 2018-04-11
Add initial gossip document

eproxus:


sylane:
Maybe there should be some mention of the peer key and how it is part of the peer URI and how it is verified ?

eproxus:
@sylane Added that in the last commit, please review.

PR #65, opened 2018-04-30
Calls are now stored in a separate tree. 

happi:
More strict definition of tree objects.
Goes with epoch PR https://github.com/aeternity/epoch/pull/1064

### DIFF
--- a/contracts/contract_state_tree.md
+++ b/contracts/contract_state_tree.md
@@ -17,25 +17,7 @@ The contract state tree contains objects:
 - Deposit (amount)
 - vm_version
 
-It is serialized as:
-```
-      CONTRACT_TYPE,
-      CONTRACT_VSN,
-      pubkey,
-      balance,
-      height,
-      nonce,
-      owner,
-      vm_version,
-      code,
-      state,
-      log,
-      active,
-      referers,
-      deposit,
-```
-
-As well as normal account fields.
+See [Contract Serialization](../serializations.md#contract) for the serialization specification.
 
 #### The contract definition
 
@@ -50,6 +32,8 @@ As well as normal account fields.
 
 The calls state tree contains all calls done in the block
 and their return values.
+
+See [Call Serialization](../serializations.md#call) for the serialization specification.
 
 - Calls & Return data
 ```
--- a/contracts/contract_state_tree.md
+++ b/contracts/contract_state_tree.md
@@ -1,9 +1,10 @@
 [back](./contracts.md)
 ## Contract state tree
 
-The contracts and their interactions reside in a contract state tree that
-is included in the root hash. The existence of a contract and the
-existence of calls can be proven.
+The contracts reside in a contract state tree that
+is included in the root hash. The existence of a contract can be proven.
+Interactions, i.e. calls and return values, are stored in their own
+state tree for one block. The existance of the call in that block can be proven.
 
 ### Contract state tree objects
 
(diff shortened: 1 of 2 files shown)


contracts/contract_state_tree.md
```diff
@@ -13,20 +14,51 @@ The contract state tree contains objects:
...
 - vm_version
 
+It is serialized as:
+```
+      CONTRACT_TYPE,
+      CONTRACT_VSN,
+      pubkey,
+      balance,
+      height,
+      nonce,
```
lucafavatella: I suggest deleting this paragraph from here and rather review https://github.com/aeternity/protocol/blob/b9b9eb2a707a18b7b976be78f3eca76e27c9c5e3/serializations.md#contract e.g. unclear (missing?) nonce there.

contracts/contract_state_tree.md
```diff
@@ -13,20 +14,51 @@ The contract state tree contains objects:
...
 
-### Pruning of contracts
+The calls state tree contains all calls done in the block
+and their return values.
+
+- Calls & Return data
+```
+      	CONTRACT_INTERACTION_TYPE,
+      	CONTRACT_INTERACTION_VSN,
+      	caller_address,
```
lucafavatella: I suggest deleting this paragraph and reviewing https://github.com/aeternity/protocol/blob/b9b9eb2a707a18b7b976be78f3eca76e27c9c5e3/serializations.md#contract-call - there are two nonces there.

contracts/contract_state_tree.md
happi: Hmm we need to describe both how a transaction is serizlized and how a state tree is serialized. Do we descibe both somewhere in protocol now?

happi: Aha, we do in that version.

contracts/contract_state_tree.md
```diff
@@ -13,20 +14,23 @@ The contract state tree contains objects:
...
 
+### Pruning of contracts
+
+## Calls state tree
 
+The calls state tree contains all calls done in the block
+and their return values.
 
-### Pruning of contracts
+See [Call Serialization](../serializations.md#call) for the serialization specification.
```
lucafavatella: I do not see `#call` anchor. Maybe `#contract-call`?

contracts/contract_state_tree.md
```diff
@@ -13,20 +14,23 @@ The contract state tree contains objects:
 - The contract log (data binary)
 - Active flag (Boolean)
 - List of refering contracts
-- Calls & Return data
 - Deposit (amount)
 - vm_version
 
-As well as normal account fields.
+See [Contract Serialization](../serializations.md#contract) for the serialization specification.
```
lucafavatella: It looks like a nonce is missing in the doc https://github.com/aeternity/epoch/blob/098fcfad68adc13fac7842103de299c20ffbc22d/apps/aecontract/src/aect_contracts.erl#L187

serializations.md
```diff
@@ -245,6 +245,7 @@ subsequent sections divided by object.
 , <contract_address> :: binary()
```
lucafavatella: Is the nonce field expected here? it looks like there are two nonces here (while code has one https://github.com/aeternity/epoch/blob/098fcfad68adc13fac7842103de299c20ffbc22d/apps/aecontract/src/aect_call.erl#L133 ).

PR #66, opened 2018-05-03
Move coinbase to header

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/157214052

### DIFF
--- a/serializations.md
+++ b/serializations.md
@@ -31,7 +31,7 @@ constructed directly as a byte array.
 | pow_evidence | 168  |
 | nonce        | 8    |
 | time         | 8    |
-
+| miner        | 32   |
 
 #### Block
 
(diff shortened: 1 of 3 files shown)


serializations.md
```diff
@@ -31,7 +31,7 @@ constructed directly as a byte array.
 | pow_evidence | 168  |
 | nonce        | 8    |
 | time         | 8    |
-
```
sennui: I am a little bit divided on that. On one hand it follows the Ethereum pattern, on the other there is more than one beneficiary (because we split the computed value 60-40).

lucafavatella: Moved to miner.

lucafavatella:
Force-pushed with rebased on current master and with 2 extra commits - one re language and one re consensus doc.

PR #67, opened 2018-05-08
Sophia documentation update

UlfNorell:


lucafavatella:
I understand this relates to https://github.com/aeternity/epoch/pull/1097

contracts/sophia.md
```diff
@@ -62,11 +62,14 @@ worked out).
 Sophia does not have arbitrary mutable state, but only a limited form of
 state associated with each contract instance.
 
-- Each contract defines a state type encapsulating its mutable state.
+- Each contract defines a type `state` encapsulating its mutable state.
+- The initial state of a contract is computed by the contract's `init`
+  function. The `init` function is *pure* and returns the initial state as its
```
lucafavatella: Can the init function call other contracts?
Can the init function call other contracts by name?
Can the init function interrogate the chain state?

contracts/sophia.md
```diff
@@ -149,32 +149,27 @@ type transaction = SpendTx(spend_tx)
...
-- `chain.timestamp` is the timestamp of the current block
+- `Call.value` is the amount of coins transferred to the contract in the call.
+- `Call.gas_price` is the gas price of the current call.
+- `Call.gas_left` is the amount of gas left for the current call.
+- `Chain.get_balance(a : address)` returns the balance of account `a`.
+- `Chain.block_hash(h)` returns the hash of the block at height `h`.
+- `Chain.block_height` is the height of the current block (i.e. the block in which the current call will be included).
+- `Chain.coinbase` is the address of the account getting the coinbase transaction of the current block.
+- `Chain.timestamp` is the timestamp of the current block.
+- `Chain.difficulty` is the difficulty of the current block.
```
lucafavatella: Target threshold (that is in the header) or difficulty (that would be a derived value)?

contracts/sophia.md
```diff
@@ -149,32 +149,27 @@ type transaction = SpendTx(spend_tx)
...
-- `self.creator` is the address of the entity that signed the contract creation
-  transaction
-- `self.address` is the address of the contract account
-- `self.balance` is the amount of coins currently in the contract account
-- `chain.height` is the height of the current block (i.e. the block in which the current call will be included)
-- `chain.timestamp` is the timestamp of the current block
+- `Call.value` is the amount of coins transferred to the contract in the call.
+- `Call.gas_price` is the gas price of the current call.
+- `Call.gas_left` is the amount of gas left for the current call.
+- `Chain.get_balance(a : address)` returns the balance of account `a`.
```
lucafavatella: I expect this is the balance in the chain state passed as input to the contract call (as other txs placed before this tx in the same block may have changed the chain state).

contracts/sophia.md
```diff
@@ -149,32 +149,27 @@ type transaction = SpendTx(spend_tx)
...
-A contract has access to variables of the same names, so that
-
-- `call.caller` is the address of the entity (possibly another contract)
+The block-chain environment available to a contract is defined in three name spaces
+`Contract`, `Call`, and `Chain`:
+
+- `Contract.creator` is the address of the entity that signed the contract creation
+  transaction.
+- `Contract.address` is the address of the contract account.
+- `Contract.balance` is the amount of coins currently in the contract account.
```
lucafavatella: I expect this is the balance in the chain state passed as input to the contract call (as other txs placed before this tx in the same block may have changed the chain state).

contracts/sophia.md
```diff
@@ -546,6 +546,15 @@ The shortest unique suffix could also be used, and the hash could be ANDed with
 Then each exported function starts with an exported entry point where it executes
 code to fetch the arguments to memory.
 
+### Contract state
+
+Before a contract call is executed the current state of the contract is loaded
+into memory encoded as described above. A pointer to the encoded state is stored at
+address 0. The Sophia compiler generates code to unpack the encoded state and
```
lucafavatella: Could you please confirm address 0 in other places - I found:
https://github.com/aeternity/protocol/pull/67/commits/84d3b0931235e394715ce762cb2fce68f2c33835#diff-9ddeae469112ac9dd44768c45511c6e9R527
https://github.com/aeternity/protocol/pull/67/commits/84d3b0931235e394715ce762cb2fce68f2c33835#diff-9ddeae469112ac9dd44768c45511c6e9R528

lucafavatella: Maybe other addresses shall be documented? (I am thinking 32 and/or 64 based on skimming https://github.com/aeternity/epoch/pull/1097 )

contracts/sophia.md
UlfNorell: At the moment, yes (except calling by name, which isn't implemented at all). Whether it should be allowed, not sure.

contracts/sophia.md
UlfNorell: It's the balance from the current state trees. The contract call itself might have updated it as well.

contracts/sophia.md
UlfNorell: Ah, well spotted. Those should be addr 32.

PR #69, opened 2018-05-08
Proper description of Sophia ABI data layout

UlfNorell:


contracts/sophia.md
```diff
@@ -484,7 +484,7 @@ and miners to pick disable transactions.
 
 ## The Sophia_01 ABI
 
-The calldata contains a tuple with function name and argument. E.g. ("main", (1,2,3))
+The calldata contains a tuple with function name and argument, e.g. `("main", (1,2,3))`.
```
lucafavatella: I see code is https://github.com/aeternity/epoch/blob/86772227ad547e05d0df21337b0f171e16ed258b/apps/aehttp/src/aehttp_dispatch_ext.erl#L452 and that argument does not need to be tuple.

happi: That off-chain call creates call data from function and argument.

PR #70, opened 2018-05-08
Describe MPT storage of contract storage state.

happi:


serializations.md
```diff
@@ -229,14 +229,31 @@ subsequent sections divided by object.
...
+The contract storage (or state) which is a key value map from (key::binary() to value::binary())
+is stored in its own subtree. The key for a contract storage value is:
+```
+<contractpubkey><16><key> :: binary()
+```
+Each value is just stored as a binary as is. If the value is the empty binary the key is pruned
+from the tree.
+
+Contracts with vm_version == 1, i.e. Sophia contracts on the AEVM stores the memory layout of the
+tate as one binary value at address 0.
```
gorillainduction: tate -> state

serializations.md
```diff
@@ -229,14 +229,31 @@ subsequent sections divided by object.
...
+from the tree.
+
+Contracts with vm_version == 1, i.e. Sophia contracts on the AEVM stores the memory layout of the
+state as one binary value at address 0.
+
+Contracts with vm_version == 2, i.e. Solidity contracts on the AEVM
+stores the VM storage map containg 256 bit binaries for both keys and
+values in the MPT. Undefined keys are treated as having the value 0
+and keys bound to the value zero are stored as the empty binary, thus
+purging them from the tree.
```
lucafavatella: It looks like contract is never deleted from state tree? If it were deleted, all of this state (i.e. all keys with specific prefix) were to be deleted too, right?

serializations.md
```diff
@@ -229,14 +229,31 @@ subsequent sections divided by object.
...
+Each value is just stored as a binary as is. If the value is the empty binary the key is pruned
+from the tree.
+
+Contracts with vm_version == 1, i.e. Sophia contracts on the AEVM stores the memory layout of the
+state as one binary value at address 0.
+
+Contracts with vm_version == 2, i.e. Solidity contracts on the AEVM
+stores the VM storage map containg 256 bit binaries for both keys and
+values in the MPT. Undefined keys are treated as having the value 0
+and keys bound to the value zero are stored as the empty binary, thus
```
lucafavatella: I found these tests https://github.com/aeternity/epoch/pull/1096/files#diff-e68d1e1a926ce9e96bbcd4b81f8668cdR393 though did not find when this defaulting and storing of special values is called.

PR #71, opened 2018-05-08
Design to be more resilient against sybil/eclipse attacks

sylane:


GOSSIP.md
```diff
@@ -2,30 +2,31 @@
...
 
-```
+'''
 aenode://pp$HryRGHJ7Ct3trkktVyVBgfhHL1J4EYSD9cScuMZDV61eSHrCZ@mynode.example.com:3015
-```
+'''
 
-A node is uniquely identified by the triplet of public key, hostname and port.
-This means one hostname can have several instanced of Epoch started at the same
+A node is uniquely identified by the triplet of public key, IP and port.
```
hanssv: This is actually wrong, or at least not according to the implementation, a node is uniquely identified by the public key only!

GOSSIP.md
```diff
@@ -34,45 +35,252 @@ also used to determine which node should keep its initiated connection open in
...
+counter and last retry time are updated. The node keeps picking peers and
+connecting until the maximum number of concurrent connections is reached.
+
+The peer retry counter and last retry time (initialized to `0` and `infinity`),
+are used to filter out peers when picking them from the pools providing
+exponential backoff. After N failed attempts, a verified peer that is not
+marked as trusted is downgraded to the unverified pool and the counter and time
+is reset; unverified peers are simply removed.
+
+Connected peers will periodically be pinged, and their connection state
```
hanssv: Might help the reader to point out that the ping is solely for chain status information - i.e. it is not needed to track connection state etc.

GOSSIP.md
```diff
@@ -34,45 +35,252 @@ also used to determine which node should keep its initiated connection open in
...
+pool.
+- The maximum number of neighboring peers sharing the same IP a single node
+can add through gossip is from 128 to 1024 depending on peer duplication in the
+unverified pool.
+
+#### Concerns (WIP)
+
+- There is always a cost for updating an unverified gossiped peer:
+
+  If a peer is not in the verified pool BUT already was in the unverified pool,
```
hanssv: I think this is a problem especially in the beginning where the total set of peers/nodes is pretty small...

I might be missing something, but the fact that we are gossiped the same peer from multiple sources should favor this peer? If we simply remove it and then add it again that effect is missing?

GOSSIP.md
```diff
@@ -34,45 +35,252 @@ also used to determine which node should keep its initiated connection open in
...
+
+Only the IP of the source node is used to select a subset of the pool because
+IP is an expensive resource while ports are comparatively cheap.
+
+See [Bucket Selection](#bucket-selection) for more details on how the buckets
+are selected from the secret and other discriminators.
+
+#### Verified Pool
+
+The verified pool is composed of 256 buckets of up to 64 peers, resulting in a
```
hanssv: This is far too many! As a comparison there are a total of ~16000 ethereum nodes at the moment...

GOSSIP.md
eproxus: But practically it is bounded by the IP/port as well, right? You can't have two nodes with different keys connected through the same socket, no? 🙂

GOSSIP.md
eproxus: Will this change at some point? There was some discussion regarding connection timeout / cleanup regarding the system tests...

GOSSIP.md
sylane: Maybe it should be reworded, the pubkey is the only "identity" of a node, but there is a single IP/port for a not to. There couldn't be multiple IP/port with the same pubkey or multiple pubkey with the same IP/port. Is that correct?

GOSSIP.md
sylane: A 2 minutes ping with retry would not be very helpfull for detecting lost connections anyway. If we wanted to speed up dead connection detection we would need a ping frequency a lot higher or a different kind of ping, maybe just sending a dummy byte and monitoring activity...

GOSSIP.md
sylane: The issue with not cleaning the references first is that we don't control duplication, we could end up with 1024 (one per bucket) references, this would go against the idea of limiting pool poisoning. Another option is to completely forget about duplication and keep the peer in the bucket it was added the first time.

GOSSIP.md
sylane: I understand, maybe 128 buckets are enough. Keep in mind that due to the way bucket are selected buckets will be filled in a linear way so the maximum should be bigger than the target.

hanssv:
https://github.com/aeternity/protocol/commit/86ed9b776598462a03b1f94853dc8def8daf6f43 is this related? Orthogonal?

eproxus:
@hanssv Would have been nice to know that commit existed 🙂 Looks like the paragraph about "neighbor selection" only dictates that we should come up with something. The rest looks unrelated...

@knarz Does this proposal sound good to you in the light of what you wrote in 86ed9b776598462a03b1f94853dc8def8daf6f43?

Finally, we should standardize a common name for this concept. In the code we use _peers_, and in `SYNC.md` we use "neighbor". This only adds confusion.

eproxus:
We should also avoid doing double work here, the #core-sync chat exists for a reason 🙂

GOSSIP.md
knarz: If we have a node with a dynamic IP address then we might end up with multiple `(IPADDR, PORT)` pairs with the same pubkey, although that should be a transient state. (Or maybe if a node has multiple interfaces and listens on them too?)

Multiple pubkeys with the same `(IPADDR, PORT)` pair should not be possible.

knarz:
This is close to what bitcoin is doing, right? Seems like a sane approach and if we could somehow merge the two documents and unite them under the P2P topic, then that'd be great.

If we want to emphasise, that they are equal, then we use peer, and if we care more about them being closely located, then neighbor. I really don't mind either but if I had to make a choice given the previous points, I might prefer peer.

GOSSIP.md
sylane: So we should remove any known peer with the same pubkey but different IP/port before adding a new peer.

sylane:
In the documentation, I tend to use `node` when talking about the service where the "action" is taking place, `peer` for any other service the `node` could know about, and `neighbor` or `neighboring peers` for the services a gossiping `peer` is sending to the `node`. In other words, the `neighbors` are the verified `peers`.

sylane:
@hanssv, @eproxus, @knarz, @ThomasArts  I investigated the issues with the design regarding the cost of adding a peer that already exists and allowing new nodes to connect to the cluster. I updated the document taking the feedback into account. Please have a look, I would appreciate any feedback. If there are no major issues, I am planning on starting to work on the implementation this week.

knarz:
I'm missing a distinction between incoming/outgoing connections, i.e. only opening a limited number (e.g. 12) of outgoing connections while allowing many more incoming ones. Other than that this seems ok to me.

Once this is merged, I'd go ahead and then try to combine this why what I've written plus simple reputation and strategies for compact blocks.

sylane:
@sh I think there is some confusion regarding incoming/outgoing connections. The idea is to group the connection in two groups, a large pool of connection to verified peers (could be outgoing or incoming) and a small pool of incoming connection from unverified peers. The main difference of these pools (besides there size) is that the unverified connection pool is a FIFO were new connection make the oldest existing one be closed. Outgoing connections are a lot safer against sybil/eclipse attack because an attacker cannot force the selection of there peers that would have made it to the verified peers. Are you saying that the node shouldn't try to connect to more than ~12 peers and all the other verified connection should come from the incoming connections from verified peers? I think that wouldn't work with this design, because to be considered verified, a peer REQUIRES an outgoing connection, if we limite the number of outgoing connection to something very small like 12, we wouldn't have much verified peers and all the incoming connection would go to the small unverified pool.

GOSSIP.md
```diff
@@ -1,78 +1,338 @@
-# Gossip
+4# Gossip
```
lucafavatella: 4?

sylane: Good catch...

lucafavatella:
I took the liberty of marking as WIP as I understand from @sylane that some changes may be required.

sylane:
@hanssv @knarz : From what I could see, the bitcoin server reference code is throwing the dice to select which pool a new peer is picked from (new/tried); there is 50% chance for each. Is it something we would want to do too, or should we always pick from the verified pool (tried)? As I understand it, this would augment the rate at which the verified pool is filled by a small margin.

GOSSIP.md
```diff
@@ -2,77 +2,323 @@
...
-A peer is identified by a URI consisting of the protocol `aenode://`, the public
-key, an `@` character, the hostname or IP number, a `:` character and the Noise
-port number the node is using. Example:
+A peer is identified by a URI consisting of the protocol 'aenode://', the public
+key, an '@' character, the hostname or IP number, a ':' character and the Noise
+port number the node is using. If the address contains a hostname, it will be
+resolved to an IP. Example:
 
-```
+'''
```
eproxus: These should still be backpacks, right? 🙂

eproxus: There are a few more places in the document that uses single quotes instead of backticks.

GOSSIP.md
```diff
@@ -2,77 +2,323 @@
...
 ```
 
-A node is uniquely identified by the triplet of public key, hostname and port.
-This means one hostname can have several instanced of Epoch started at the same
-time listening on different ports. In the case of several unique public keys for
-the same hostname and port, only one can work. This will be the public key
-corresponding to the private key that is used for the Noise protocol listener
-associated with the hostname and port.
+A node is uniquely identified by its public key, and it can only have one IP and
+port at the same time. This means one IP can have several instanced of Epoch
```
lucafavatella: instances?

GOSSIP.md
```diff
@@ -2,77 +2,323 @@
...
+started at the same time listening on different ports if they have different
+public keys. This key corresponds to the private key that is used for the Noise
+protocol listener associated with the IP and port.
 
 Apart from being used in encryption via the Noise protocol, the public key is
 also used to determine which node should keep its initiated connection open in
 case two connections are opened (see [Gossiping of New Peers](#gossiping-of-new-
 peers)).
```
lucafavatella: broken link though not touched by this PR

GOSSIP.md
```diff
@@ -2,77 +2,323 @@
...
+Peers are grouped by the 16 most significant bits of a peer IP (\16 mask).
+This group is used when selecting a bucket and establishing the node outbound
+connections to prevent nodes from being connected only to local nodes; this
+make block forwarding more effective. NOTE: If all the nodes are in the same
+group, they may not reach their maximum number of outbound connection and this
+should probably be disabled.
+
+### Connections
+
+There is two pool of connections, inbound connection and outbound connections.
```
lucafavatella: are?

GOSSIP.md
```diff
@@ -2,77 +2,323 @@
...
+
+There is a hard limit on the maximum number of outbound connection
+(default to 10).
+
+There is a soft limit on the maximum of inbound connections (default to 100).
+When it is reached, the node will still accept inbound connections, but they will
+be closed right after responding to the first ping. This is a soft limit to allow
+nodes to join the cluster even if the trusted nodes inbound limit has been
+reached. In addition, this prevent a node, in particular a trusted node, to
+have so much inbound connections that it cannot reach its number of outbound
```
lucafavatella: many?

knarz:
I think taking from both pools is good, but this also seems like something, that would be great for simulations.

GOSSIP.md
```diff
@@ -2,77 +2,323 @@
...
+##### Bucket Selection
+
+The bucket selection is done by hashing the secret with the discriminator and
+use the result as an integer; this integer modulo is used to restrict the
+subset.
+
+For the unverified bucket selection:
+
+```Erlang
+<<N1:160>> = crypto:hash(sha, <<Secret/binary, PeerGroup/binary>>).
```
lucafavatella: SHA-1? (Just confirming.)

sylane: Yes, it is.

GOSSIP.md
sylane: What do you mean?  Replace `so much inbound connection` to `many inbound connections`? I feel this doesn't mean the same, but I may be wrong.

knarz: »In addition, this prevents a node, in particular a trusted one, to end up with too many inbound connections. This could prevent it from establishing further outbound connections, due to the global maximum«

This is what you mean, I assume?

sylane: This is not really the issue. The problem is that because all the nodes connect to the trusted nodes, it is already connected to all pooled peers so it will not connect to any. Another option would be to force outbound connection even though a node already have an inbound connection for the peer but this doesn't sound like a good idea.

PR #79, opened 2018-05-15
Draft: Detail contract referers order in state tree

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/157593451

May not be in line with code.


lucafavatella:
I close this - it is referred by https://www.pivotaltracker.com/n/projects/2159793/stories/157593451

PR #80, opened 2018-05-15
Remove unneeded height from account

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/157308030

Not ready to be merged until related code is ready.


lucafavatella:
Depends on #82

PR #82, opened 2018-05-15
Update to new contract serialization with balance in account tree.

happi:
The new serialization introduced in https://github.com/aeternity/epoch/pull/1131

### DIFF
--- a/serializations.md
+++ b/serializations.md
@@ -231,11 +231,7 @@ subsequent sections divided by object.
 
 #### Contract
 ```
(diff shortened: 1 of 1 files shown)


serializations.md
```diff
@@ -245,6 +241,7 @@ subsequent sections divided by object.
 ]
 ```
 
+The balance of the accout is stored in the account state tree.
```
gorillainduction: accout -> account

PR #89, opened 2018-05-21
Pt 157695429 remove 'from' field in close mutual

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/157695429


gorillainduction:
@lucafavatella Integers are defined to be unsigned, and as they are RLP encoded, the serialization will handle (almost) any size. RLP also handles different size binaries. The semantics of the values are up to higher level documentation to describe.

PR #91, opened 2018-05-23
updated for name hash resolvement

thepiwo:


thepiwo:
so this seems to be good as described and discussed @lucafavatella can we merge?

lucafavatella:
> can we merge?

@thepiwo Please let us wait until corresponding code is ready - so we can merge spec / doc and code in same release.

evdoks:
Was implemented in a different way. Closing.

PR #92, opened 2018-05-23
Introduce serialization format for key and micro blocks

jur0:


serializations.md
```diff
@@ -15,29 +15,49 @@ user API.
...
 
-#### Block header
+There are two types of blocks:
 
-For block headers all fields sizes are statically known and can be
+* key block - does not contain a list of transactions, therefore, it is
+not divided into a header and a list of transactions;
+* micro block - contains a header and a list of transactions.
+
+#### Key block
```
lucafavatella: I know it was not introduced by this PR though it is strange that heading passes from `##` to `####` without `###`.

serializations.md
```diff
@@ -121,6 +141,8 @@ subsequent sections divided by object.
 | Channel settle transaction | 56 |
 | Channel off chain transaction | 57 |
 | Channel | 58 |
+| Key block | 100 |
```
lucafavatella: Not needed as it only makes sense for serialization as `S(O) = rlp([tag(O), vsn(O)] ++ fields(O))` while serialization is defined as byte array.

serializations.md
```diff
@@ -121,6 +141,8 @@ subsequent sections divided by object.
...
+| Micro block | 101 |
```
lucafavatella: Not needed as it only makes sense for serialization as `S(O) = rlp([tag(O), vsn(O)] ++ fields(O))` while serialization is defined as byte array.

serializations.md
```diff
@@ -15,29 +15,49 @@ user API.
...
-| miner        | 32   |
-
-#### Block
-
-The only difference between a block and its header is the list of
+| version    | 8    |
+| prev_hash  | 32   |
+| state_hash | 32   |
+| txs_hash   | 32   |
+| miner_hash | 32   |
```
lucafavatella: How is `miner_hash` in microblock different from `miner` in key block? If same thing, naming seems inconsistent.

lucafavatella: I understand miner is not strictly needed in microblock as it can be inferred from lastest keyblock linked (potentially transitively) by previous hash: is it kept for easier signature verification?

jur0: `miner_hash` is a hash of a `miner` - the public key of a miner.

jur0: `miner_hash` is not needed, but for now it's included in the micro blocks as it makes it possible to figure out what key block a micro block belongs to. This info could be retrieved from `aec_chain_state`, although, it's not implemented yet.

serializations.md
lucafavatella: See also https://github.com/aeternity/epoch/pull/1202#discussion_r191372961 for more context.

serializations.md
```diff
@@ -15,31 +15,56 @@ user API.
...
+| --- | --- |
+| version    | 8    |
+| height     | 8    |
+| prev_hash  | 32   |
+| state_hash | 32   |
+| miner      | 32   |
+| target     | 8    |
+| pow        | 168  |
+| nonce      | 8    |
+| time       | 8    |
```
lucafavatella: I see in line with https://github.com/aeternity/epoch/pull/1293/files#diff-3422fb1f1624d1021226c01eed4d0cd0R140
Ok.

serializations.md
```diff
@@ -15,31 +15,56 @@ user API.
...
-transaction root hash (txs_hash). The block does not currently have a
-separate binary serialization form since the block hash is computed
-from the block header.
+| version    | 8    |
+| height     | 8    |
+| prev_hash  | 32   |
+| state_hash | 32   |
+| txs_hash   | 32   |
+| key_hash   | 32   |
+| time       | 8    |
```
lucafavatella: I see in line with https://github.com/aeternity/epoch/pull/1293/files#diff-3422fb1f1624d1021226c01eed4d0cd0R152
Ok.

serializations.md
```diff
@@ -15,31 +15,56 @@ user API.
...
+
+### Micro block
+
+The only difference between a micro block and its header is a list of
+transactions and a signature. The transactions are captured in the
+header by the transaction root hash (`txs_hash`). The block does not
+currently have a separate binary serialization form since the block
+hash is computed from the block header. The signature comes after the
```
lucafavatella: As block does not impact consensus and this doc is about serializations, the sentence "The signature comes after the list of transactions ..." and the detail on the signature could be deleted.

PR #95, opened 2018-05-24
Add a simple rent model and lock for transfers to AENS

knarz:
This change introduces a simple, fixed rent model to the naming system.

In addition, a locking period is attached to a name transfer, which aims to give users of that name a grace period to adjust to owner changes. (Are two weeks enough?)

### DIFF
--- a/AENS.md
+++ b/AENS.md
@@ -52,9 +52,9 @@ the general public can start claiming names.
 
 ## Mechanisms
 
-It is unclear what a good mechanism for a naming system would look
-like. If we imagine two actors both being interested in the same name,
-what would a »fair« solution be to resolve this?
+It is unclear what a good distribution mechanism for a naming system would look
+like. If we imagine two actors both being interested in the same name, what
+would a »fair« solution be to resolve this?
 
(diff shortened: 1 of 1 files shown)


AENS.md
```diff
@@ -268,27 +268,30 @@ commitment := Hash(NameHash(name) + name_salt)
...
-`pre-claim` transaction containing a commitment to the name and nonce.
+The `claim` transaction MUST be signed by the same private key as a `pre-claim`
+transaction containing a commitment to the name and nonce.
 
 A `claim` transaction MUST NOT be in included in the same block as its
 `pre-claim`.
 
-Note that the fee here is a distinct fee from the normal transaction fee.
+`rent` is the reoccurring fee that has to be paid for names. Our first rent
+model is a naïve one, that just uses a flat fee. Our initial target is 4 aeons
```
velzevur: Do we use *aeon* as a name of the currency? Don't we rather use _AE tokens_, instead? Main concern is that _Aeon_ is already used as a currency name...
Same comment for the line bellow.

PR #99, opened 2018-05-24
Illustrate transaction unpacking and signing

mpowaga:
https://github.com/aeternity/aepp-sdk-js/pull/37 needs to be merged first.

Things that could be improved:

- Explain how `unpack` and `sign` tools can be used (maybe `npm install -g @aeternity/aepp-sdk`?)
- Use hex keys instead of `--file` option so that user following the documentation will get the same results
- Maybe remove `unpack` step when second-party is to decode the same transaction (since it gives the same result)


epoch/api/channels_api_usage.md
```diff
@@ -75,7 +75,12 @@ in the chain, and the others are metadata used for the connection itself.
   | ttl | 1000 |
 
   The `initiator` will be connecting to the `responder` on localhost:1234
-  We will be using the tool [wscat](https://github.com/websockets/wscat)
+  We will be using following tools
+
+  * [wscat](https://github.com/websockets/wscat)
+  * [unpack](https://github.com/aeternity/aepp-sdk-js/blob/develop/examples/sign.js)
```
velzevur: broken link

epoch/api/channels_api_usage.md
```diff
@@ -75,7 +75,12 @@ in the chain, and the others are metadata used for the connection itself.
...
+  * [sign](https://github.com/aeternity/aepp-sdk-js/blob/develop/examples/sign.js)
```
velzevur: broken link

velzevur:
Looks good to me

mpowaga:
I've updated example messages so they correspond to channel parameters and used hex keys so users can get the same results when signing/unpacking txs.

/cc @velzevur @uwiger

velzevur:
Since this has to be updated according to latest changes in the protocol, I've added a WIP label

PR #104, opened 2018-05-28
State channel parties can agree upon any final share as long as equal…

ThomasArts:
… to total in channel

### DIFF
--- a/channels/ON-CHAIN.md
+++ b/channels/ON-CHAIN.md
@@ -159,7 +159,7 @@ This transaction MUST have valid signatures of all involved parties.
 After this transaction has been included in a block, the channel MUST be
 considered closed and allow no further modifications.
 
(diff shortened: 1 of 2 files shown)


channels/ON-CHAIN.md
```diff
@@ -176,6 +159,8 @@ This transaction MUST have valid signatures of all involved parties.
 After this transaction has been included in a block, the channel MUST be
 considered closed and allow no further modifications.
 
+`channel initiator_amount + responder_amount ==
```
velzevur: The internal distribution of tokens inside the channel is not known on-chain so we don't really have the notion of `channel initiator_amount + responder_amount`. In the light of withdrawals and deposits, this certainly can not be the channel create balances...
Probably `channel total amount` could fit better?

PR #106, opened 2018-05-30
Add Proof of inclusion serialization

gorillainduction:
https://github.com/aeternity/epoch/pull/1202
https://www.pivotaltracker.com/n/projects/2124891/stories/157873700

### DIFF
--- a/serializations.md
+++ b/serializations.md
@@ -121,6 +121,7 @@ subsequent sections divided by object.
 | Channel settle transaction | 56 |
 | Channel off chain transaction | 57 |
 | Channel | 58 |
+| POI | 60 |
 
 #### Accounts
 ```
@@ -480,3 +481,23 @@ The channel offchain transaction is not included directly in the transaction tre
 , <closes_at>       , int()
 ]
 ```
+
+#### Proof of inclusion on state trees (POI)
+```
+[ {<accounts>  , [{binary(), [{binary(), [binary()]}]}]}
+, {<calls>     , [{binary(), [{binary(), [binary()]}]}]}
+, {<channels>  , [{binary(), [{binary(), [binary()]}]}]}
+, {<contracts> , [{binary(), [{binary(), [binary()]}]}]}
+, {<ns>        , [{binary(), [{binary(), [binary()]}]}]}
+, {<oracles>   , [{binary(), [{binary(), [binary()]}]}]}
+]
+```
+
+Each field is the representation of a subtree to the top level tree. The serialization should be interpreted as
+```
+{<tree>, [<proof_of_inclusion>]}
+where
+<proof_of_inclusion> := [{<root_hash>, [{<mpt_hash>, <mpt_value>}]}]
+```
+where `[{<mpt_hash>, <mpt_value>}]` is the sorted list of Merkel Patricia Tree nodes in the proof. If the subtree is empty,
+the serialization is just `[]`, and otherwise it is a list of one element `[<proof_of_inclusion>]`.
--- a/serializations.md
+++ b/serializations.md
@@ -366,59 +366,59 @@ purging them from the tree.
 
 #### Channel create transaction
 ```
-[ <initiator>        , binary()
-, <initiator_amount> , int()
-, <responder>        , binary()
-, <responder_amount> , int()
-, <channel_reserve>  , int()
-, <lock_period>      , int()
-, <ttl>              , int()
-, <fee>              , int()
-, <nonce>            , int()
+[ <initiator>        :: binary()
+, <initiator_amount> :: int()
+, <responder>        :: binary()
+, <responder_amount> :: int()
+, <channel_reserve>  :: int()
+, <lock_period>      :: int()
+, <ttl>              :: int()
+, <fee>              :: int()
+, <nonce>            :: int()
 ]
 ```
 
 #### Channel deposit transaction
 ```
-[ <channel_id> , binary()
-, <from>       , binary()
-, <amount>     , int()
-, <ttl>        , int()
-, <fee>        , int()
-, <nonce>      , int()
+[ <channel_id> :: binary()
+, <from>       :: binary()
+, <amount>     :: int()
+, <ttl>        :: int()
+, <fee>        :: int()
+, <nonce>      :: int()
 ]
 ```
 
 #### Channel withdraw transaction
 ```
-[ <channel_id> , binary()
-, <to>         , binary()
-, <amount>     , int()
-, <ttl>        , int()
-, <fee>        , int()
-, <nonce>      , int()
+[ <channel_id> :: binary()
+, <to>         :: binary()
+, <amount>     :: int()
+, <ttl>        :: int()
+, <fee>        :: int()
+, <nonce>      :: int()
 ]
 ```
 
 #### Channel close mutual transaction
 ```
-[ <channel_id>      , binary()
-, <initiator_amount>, int()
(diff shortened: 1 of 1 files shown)


serializations.md
```diff
@@ -121,6 +121,7 @@ subsequent sections divided by object.
 | Channel settle transaction | 56 |
 | Channel off chain transaction | 57 |
 | Channel | 58 |
+| POI | 60 |
```
lucafavatella: I see documented because of https://github.com/aeternity/epoch/pull/1202#discussion_r191644134
Ok.

serializations.md
```diff
@@ -480,3 +481,23 @@ The channel offchain transaction is not included directly in the transaction tre
...
+]
+```
+
+Each field is the representation of a subtree to the top level tree. The serialization should be interpreted as
+```
+{<tree>, [<proof_of_inclusion>]}
+where
+<proof_of_inclusion> := [{<root_hash>, [{<mpt_hash>, <mpt_value>}]}]
+```
+where `[{<mpt_hash>, <mpt_value>}]` is the sorted list of Merkel Patricia Tree nodes in the proof. If the subtree is empty,
```
lucafavatella: Can you please elaborate on the sorting? It appears to me [serialization is not sorted](https://github.com/aeternity/epoch/pull/1202/files#diff-033dadbc082ac1d0414b95ede7462cefR132) and [the verification function does not rely on sorting of `ProofDB`](https://github.com/aeternity/epoch/blob/6a320f1f9b75961145d6d7c77e761e0d3d096b3d/apps/aeutils/src/aeu_mp_trees.erl#L457-L460). Also, if proof of inclusion is "only" in a transaction (i.e. not in a state tree), for as long as all the signers sign the same binary, sorting is not strictly needed.

serializations.md
```diff
@@ -480,3 +481,23 @@ The channel offchain transaction is not included directly in the transaction tre
...
+the serialization is just `[]`, and otherwise it is a list of one element `[<proof_of_inclusion>]`.
```
lucafavatella: I understand this sentence is clarifying length of list `[<proof_of_inclusion>]` in `{<tree>, [<proof_of_inclusion>]}`.

A. Just confirming: `[]` is not a proof of absence.

B. I understand that "otherwise it is a list of one element `[<proof_of_inclusion>]`" means that this serialization can carry at most one value per subtree e.g.: one account; or one account plus one channel.

C. This implies that a POI can be serialized with no value in any subtree. Useless but legit - ok.

D. I see absent the hash of the top state tree (including accounts, channels, etc.) i.e. the one that is in block header. I think this is not sufficient, as this requires user to have not only the block headers but also the root hash of each subtree (accounts, channels, etc.) for each header.

serializations.md
```diff
@@ -480,3 +481,23 @@ The channel offchain transaction is not included directly in the transaction tre
 , <closes_at>       , int()
 ]
 ```
+
+#### Proof of inclusion on state trees (POI)
+```
+[ {<accounts>  , [{binary(), [{binary(), [binary()]}]}]}
```
lucafavatella: The division of (normative) type (here) and (informative) label (below), plus usage of `,` instead of documented `::`, makes it difficult to parse&match for the reader. May I suggest unifying the two:
```
[ {<accounts>  :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), <mpt_value> :: [binary()]}]}]}
, {<calls>     :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), <mpt_value> :: [binary()]}]}]}
, {<channels>  :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), <mpt_value> :: [binary()]}]}]}
, {<contracts> :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), <mpt_value> :: [binary()]}]}]}
, {<ns>        :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), <mpt_value> :: [binary()]}]}]}
, {<oracles>   :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), <mpt_value> :: [binary()]}]}]}
]
```

serializations.md
lucafavatella: The serialization of `<mpt_value>`, that I understand being `[binary()]`, may benefit to be further specified. Though I consider it optional for this PR.

gorillainduction: The serialization is [sorted](http://erlang.org/doc/man/gb_trees.html#to_list-1) and the deserialization used to verify this, but this has been lost apparently. It is nice to have a canonical serialization when possible, and the sorting makes the serialization canonical.

serializations.md
gorillainduction: A. This is correct.

B. No. The proof of inclusion for one tree can carry any amount of objects in the tree. Basically, the proof is a pruned node db with enough information for any number of paths in the tree.

C. Yes.

D. The root hash (from the header) can be [calculated](https://github.com/aeternity/epoch/pull/1202/files#diff-29b0252c569ea38faaa5fed4b9820676R301) from the root hashes of the subtrees. The actual root hash is omitted on purpose since it can be calculated. Analogous with the blocks (and headers) not having their own hashes in the representation.
The proof of inclusion always contains all subtrees, since this is needed to verify the root hash.

serializations.md
lucafavatella: Thanks for the link, I had missed that.

serializations.md
lucafavatella: >  The proof of inclusion always contains all subtrees, since this is needed to verify the root hash.

This is the only bit unclear to me.

serializations.md
```diff
@@ -455,28 +456,41 @@ The channel offchain transaction is not included directly in the transaction tre
...
+[ {<accounts>  :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), [<mpt_value> :: binary()]}]}]}
+, {<calls>     :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), [<mpt_value> :: binary()]}]}]}
+, {<channels>  :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), [<mpt_value> :: binary()]}]}]}
+, {<contracts> :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), [<mpt_value> :: binary()]}]}]}
+, {<ns>        :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), [<mpt_value> :: binary()]}]}]}
+, {<oracles>   :: [<proof_of_inclusion> :: {<root_hash> :: binary(), [{<mpt_hash> :: binary(), [<mpt_value> :: binary()]}]}]}
 ]
 ```
+NOTE: `[{<mpt_hash>, <mpt_value>}]` is the sorted list of Merkel Patricia Tree nodes in the proof. If the subtree is empty,
+the serialization is just `[]`, and otherwise it is a list of one element `[<proof_of_inclusion>]`.
```
lucafavatella: Ah I understood now:
* Each tree must be present;
* Each tree must have exactly one `<proof_of_inclusion>` item.

hence the root hash of each of the trees is always present hence the state hash of the corresponding block header can be computed.

PR #107, opened 2018-05-30
Adding TTL to transactions

hanssv:


hanssv:
I did go through all `epoch/api` I think. Will fix serializations.md

lucafavatella:
I see in scope of https://www.pivotaltracker.com/story/show/153897559

PR #111, opened 2018-06-01
Describe contract oracle interface, new Sophia types,  new contract APIs and how call correlates to the create or call tx

lucafavatella:


contracts/contract_state_tree.md
```diff
@@ -59,4 +59,6 @@ The calls state tree contains all calls done in the block and their
 return values. The calls state tree is pruned in each block so that
 the latest block only contains the calls in that block.
 
+The key is `<contract_address><caller_address><caller_nonce><contract_address>`.
```
lucafavatella: Is this expected to contain contract address twice?

lucafavatella: See https://github.com/aeternity/epoch/blob/c8050fea2b033b9448d1e907befcaccfa17b24ac/apps/aecontract/src/aect_call_state_tree.erl#L56-L58

contracts/contract_state_tree.md
```diff
@@ -62,3 +62,7 @@ the latest block only contains the calls in that block.
 The key is `<contract_address><caller_address><caller_nonce><contract_address>`.
 
 See [Call Serialization](../serializations.md#contract-call) for the serialization specification.
+
+Each call includes the caller address and caller nonce that, together,
+correlate the call with the transaction in the block that did the
```
lucafavatella: This comes from investigation of https://www.pivotaltracker.com/story/show/154664693
for which I am considering to determine the gas fees to give to the miner
by exploiting the `gas_used` already in each contract call object in the state tree of the block.
But, as the call object does not include the gas price, I am checking if possible to correlate the call object to the create tx or call tx in the same block in order to retrieve the gas price from that. And it seems possible correlating on caller address and caller nonce.

Other option would be to duplicate the gas price in the call object.

Other option is attempting to identify further alternatives.

lucafavatella: > I am checking if possible to correlate the call object to the create tx or call tx in the same block in order to retrieve the gas price from that. And it seems possible correlating on caller address and caller nonce.

It seems [a contract can call another contract](https://github.com/aeternity/epoch/blob/c8050fea2b033b9448d1e907befcaccfa17b24ac/apps/aecore/src/aec_vm_chain.erl#L113) that seems to create call object with caller address and nonce of the contract (not double checked). Also maybe [a contract may create another contract](https://github.com/aeternity/epoch/blob/c8050fea2b033b9448d1e907befcaccfa17b24ac/apps/aevm/src/aevm_eeevm.erl#L1458) (via `aect_create_tx`?) potentially creating call object (not double checked). Hence I am not sure (twice) it is possible to correlate call object to create or call tx.

lucafavatella: I see the code implements [correlation from contract call tx (of known hash) to contract call object](https://github.com/aeternity/epoch/blob/3e5e28a1452e8b7e17af9f940381c72b8d690d4d/apps/aehttp/src/aehttp_helpers.erl#L182-L189) (implemented by https://github.com/aeternity/epoch/pull/1185 ).

Not sure about correlation - if any - from contract create tx to contract call object.
Not sure about correlation - if any - from contract created from contract call (if any) to contract call object.

Cc @happi

happi: Call objects are only created for top-level (tx) calls. The top level call will include all gas used by any recursive calls.

contracts/aevm.md
```diff
@@ -14,6 +14,30 @@ transactions can call the contract directly. When all other contracts
 referring to a disabled contract are fully disabled the contract is
 disabled.
 
+## Æternity primitive operations
+
+Interaction with the first class objects of the Æternity chain (e.g.
+oracles, names, and state channels) is done through calls to a
+built in contract at address 0.
+
+The first argument in the call specifies witch primop to call.
```
lucafavatella: which

contracts/aevm.md
```diff
@@ -2,18 +2,43 @@
 
 The AEVM is a version of the EVM: https://github.com/ethereum/yellowpaper
 
-The AEVM is emulated inside the Ethereum node an no marshaling of
+The AEVM can be emulated inside the Æternity node an no marshaling of
```
lucafavatella: s/an/and

contracts/aevm.md
```diff
@@ -2,18 +2,43 @@
...
+Interaction with the first class objects of the Æternity chain (e.g.
+oracles, names, and state channels) is done through calls to a
+built in contract at address 0.
+
+The first argument in the call specifies witch primop to call.
+The following arguments are encoded as Sophia data.
+
+| OpNo | Name                |             Arguments | Return value    |
+| ---- | ------------------- | --------------------- | --------------- |
+|    1 | Spend               | `Recipient : address` | `Nil`           |
```
lucafavatella: I see opno in line with [code](https://github.com/aeternity/epoch/blob/3e5e28a1452e8b7e17af9f940381c72b8d690d4d/apps/aebytecode/include/aeb_opcodes.hrl#L151-L162). Ok.

contracts/aevm.md
```diff
@@ -2,18 +2,43 @@
...
 set in a disabled state where neither new contracts nor call
 transactions can call the contract directly. When all other contracts
 referring to a disabled contract are fully disabled the contract is
 disabled.
 
+## Æternity primitive operations
+
+Interaction with the first class objects of the Æternity chain (e.g.
+oracles, names, and state channels) is done through calls to a
+built in contract at address 0.
```
velzevur: built-in

contracts/aevm.md
```diff
@@ -2,18 +2,43 @@
...
+Interaction with the first class objects of the Æternity chain (e.g.
+oracles, names, and state channels) is done through calls to a
+built-in contract at address 0.
+
+The first argument in the call specifies which primop to call.
+The following arguments are encoded as Sophia data.
+
+| OpNo | Name                |             Arguments | Return value    |
+| ---- | ------------------- | --------------------- | --------------- |
+|    1 | Spend               | `Recipient : address` | `Nil`           |
```
lucafavatella: I see inline with spend [code](https://github.com/aeternity/epoch/blob/3e5e28a1452e8b7e17af9f940381c72b8d690d4d/apps/aevm/src/aevm_ae_primops.erl#L24). Ok.

contracts/sophia.md
```diff
@@ -83,25 +83,24 @@ fields as immutable?
...
+| uint       | A 256 bit integer               | ```42```
+| int        | A 256 bit 2-complement integer  | ```-1```
+| address    | A 256 bit number given as a hex | ```ff00```
+| bool       | A Boolean                       | ```true```
+| string     | An array of bytes               | ```"Foo"```
+| list       | A homogeneous immutable singly linked list. | ```[1, 2, 3]```
+| tuple      | An ordered heterogeneous array   | ```(42, "Foo", true)```
+| record     | An immutable key value store with fixed key names and typed values | ``` type balance = { owner: address, value: uint } ```
+| map        | An immutable key value store with dynamic mapping of keys of one type to values of one type | ```type accounts = map(string, address)```
+| option('a) | An optional value either None or Some('a) | ```Some(42)```
```
lucafavatella: I see this delivers https://www.pivotaltracker.com/story/show/158139745

contracts/aevm.md
```diff
@@ -2,18 +2,43 @@
...
+The AEVM provides an number of primitive operations that can be
+used to interact with the chain.
+
 The SELFDESTRUCT instruction is not immediate, instead the contract is
 set in a disabled state where neither new contracts nor call
 transactions can call the contract directly. When all other contracts
 referring to a disabled contract are fully disabled the contract is
 disabled.
 
+## Æternity primitive operations
```
lucafavatella: I see this delivers https://www.pivotaltracker.com/story/show/158023467

epoch/api/contract_api_usage.md
```diff
@@ -18,7 +18,7 @@ For an up to date list of all HTTP API endpoints and their arguments see
 An epoch node provides some utility functions to help you create contract transactions.
 
 
-### Contract Compile 
+### Contract Compile
```
lucafavatella: I see updates to this file deliver https://www.pivotaltracker.com/story/show/157839891

PR #112, opened 2018-06-05
Enrich contract call state tree object with gas price

lucafavatella:
Spec update for proposal at https://www.pivotaltracker.com/story/show/154664693/comments/190227856


lucafavatella:
Labelling as WIP as not targeted for this release, and easing filtering.

PR #117, opened 2018-06-12
Describe serializations of identifiers

gorillainduction:
* Link serializations.md from README.md
* Remove the id from some objects

https://www.pivotaltracker.com/n/projects/2124891/stories/158071477

### DIFF
--- a/serializations.md
+++ b/serializations.md
@@ -51,6 +51,30 @@ domain (e.g., `[int()]` is a list of integers). We use `++` as the
 list concatenation operator. We also use the `bool()` type as the
 special case of the integers `0` and `1` used as the boolean type.
 
+### The id() type
+
+The special type `id()` is a `binary()` denoting a tagged binary
+(e.g., hash or public key) referring to an object in the chain (e.g.,
+account, oracle) . The first byte is a tag denoting which type of
+identifier, and the remaining 32 bytes is the identifier itself. This
+is used to distinguish between identifiers where there can be
+ambiguity.
+
+| Id tag | Identifier type |
+| ---    | ---             |
+| 1      | account         |
+| 2      | name            |
+| 3      | commitment      |
+| 4      | oracle          |
+| 5      | contract        |
+| 6      | channel         |
+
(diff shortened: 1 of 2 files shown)


serializations.md
```diff
@@ -312,18 +335,17 @@ purging them from the tree.
 
 #### Name service commitment
 ```
-[ <hash>    :: binary()
```
lucafavatella: If name removed from body returned from user API, [intended usage may need update here](https://github.com/aeternity/protocol/blob/6614f009a3d033e3f8aec6a381d16def766f7eaa/epoch/api/naming_system_api_usage.md#update).

serializations.md
```diff
@@ -482,9 +504,8 @@ The channel offchain transaction is not included directly in the transaction tre
 
 #### Channel
 ```
-[ <id>               :: binary()
```
lucafavatella: Update required [here](https://github.com/aeternity/protocol/blob/6614f009a3d033e3f8aec6a381d16def766f7eaa/channels/ON-CHAIN.md#channel-state-tree) - and/or link to here.

lucafavatella: Not sure whether [off-chain channel state update](https://github.com/aeternity/protocol/blob/6614f009a3d033e3f8aec6a381d16def766f7eaa/epoch/api/channels_api_usage.md#channel-state) requires update. This would also depend on how epoch code changes contextual with this PR are performed (e.g. usage of PoI in same PR? cc @velzevur ).

serializations.md
gorillainduction: It is just the serialization that is removed. The API objects look the same.

serializations.md
gorillainduction: Again, just the serialization. Not the otherwise visible format.

gorillainduction: I believe that the actual hashes in the intended usage shouldn't be used anyway. They depend on the public keys etc. If a reader decodes the example hashes and finds that the protocol has changed, this is probably fine. The structure of the messages remain the same.

PR #121, opened 2018-06-18
Documentation of maps

UlfNorell:
For [PT-158183507](https://www.pivotaltracker.com/story/show/158183507).

See also https://github.com/aeternity/epoch/pull/1256

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -122,6 +122,82 @@ function root(t : tree('a)) : option('a) =
     Bin(_, v, _) => Some(v)
 ```
 
+### Maps and records
+
+A Sophia record type is given by a fixed set of fields with associated,
+possibly different, types. For instance
+```
+  type account = { name    : string,
+                   balance : int,
+                   history : list(transaction) }
+```
+
+Maps, on the other hand, can contain an arbitrary number of key-value bindings,
+but of a fixed type. The type of maps with keys of type `'k` and values of type
+`'v` is written `map('k, 'v)`. Key types are restricted to atomic types (`int`,
+`address`, `bool`, and `string`).
+
+#### Constructing maps and records
+
+A value of record type is constructed by giving a value for each of the fields.
+For the example above,
+```
+  function new_account(name) =
+    {name = name, balance = 0, history = []}
+```
+Maps are constructed similarly, with keys enclosed in square brackets
+```
+  function example_map() : map(string, int) =
+    {["key1"] = 1, ["key2"] = 2}
+```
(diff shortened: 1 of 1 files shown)


contracts/sophia.md
```diff
@@ -122,6 +122,82 @@ function root(t : tree('a)) : option('a) =
...
+```
+  type account = { name    : string,
+                   balance : int,
+                   history : list(transaction) }
+```
+
+Maps, on the other hand, can contain an arbitrary number of key-value bindings,
+but of a fixed type. The type of maps with keys of type `'k` and values of type
+`'v` is written `map('k, 'v)`. Key types are restricted to atomic types (`int`,
+`address`, `bool`, and `string`).
```
lucafavatella: uint?

contracts/sophia.md
```diff
@@ -122,6 +122,82 @@ function root(t : tree('a)) : option('a) =
...
+```
+Looking up a non-existing key in a map results in the contract failing. See
+`Map.member` and `Map.lookup` below for safe lookups.
+
+#### Updating a value
+
+Record field updates are written `r{f = v}`. This creates a new record value
+which is the same as `r`, but with the value of the field `f` replaced by `v`.
+Similarly, `m{[k] = v}` constructs a map with the same values as `m` except
+that `k` maps to `v`. It makes no difference if `m` has a mapping for `k` or
```
lucafavatella: So if user would like to avoid new key by mistake/typo, he/she shall define auxillary function like "map update" asserting that key is member before updating value. Ok.

contracts/sophia.md
```diff
@@ -122,6 +122,82 @@ function root(t : tree('a)) : option('a) =
...
+
+#### Builtin functions on maps
+
+The following builtin functions are defined on maps:
+
+```
+  Map.lookup(k : 'k, m : map('k, 'v)) : option('v)
+  Map.member(k : 'k, m : map('k, 'v)) : bool
+  Map.delete(k : 'k, m : map('k, 'v)) : map('k, 'v)
+  Map.size(m : map('k, 'v)) : int
```
lucafavatella: uint?

contracts/sophia.md
UlfNorell: We only have `int` at the moment. I'll make sure to update it when we add `uint`s.

contracts/sophia.md
UlfNorell: Yes, or use the `@`-construct, which does require the key to be present: `m{ [k] @ _ = v }`.

PR #125, opened 2018-06-20
Clarify that contract store object has no tag or version

lucafavatella:
Prepared while reviewing https://github.com/aeternity/epoch/pull/1271 though it is independent from it.


lucafavatella:
I will open new PR.

PR #126, opened 2018-06-21
Add channel's intended usage for deposit/withdrawal

velzevur:
Changes corresponding to aeternity/epoch#1276

PR #133, opened 2018-06-27
Add channel's intended usage for deposit/withdrawal

velzevur:
This is a squashed version of #126

### DIFF
--- a/epoch/api/channels_api_usage.md
+++ b/epoch/api/channels_api_usage.md
@@ -30,13 +30,16 @@ but are not necessarily part of the channel's life cycle.
 
 * [Generic messages](#generic-messages)
 
+* [Deposit](#deposit-events)
+
+* [Withdrawal](#withdraw-events)
+
 Only steps 1 and 3 require chain interactions, step 2 is off-chain.
 
 ### HTTP requests
 There are two types of HTTP requests:
-* Amount-modifying ones - [deposit](#deposit-transaction) and [withdrawal](#withdrawal-transaction)
-* Channel-closing ones - [solo close](#solo-close-transaction),
-  [slash](#slash-transaction) and [settle](#settle-transaction)
+* Amount-modifying ones - deposit and withdrawal
+* Channel-closing ones - solo close, slash and settle
 
 ## Channel open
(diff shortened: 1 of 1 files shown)


epoch/api/channels_api_usage.md
```diff
@@ -518,3 +521,268 @@ details:
...
+closing a channel on-chain, the closing balances of the participants are
+checked against this balance. Under no circumstances can the sum of the closing balances
+be greater than the total balance on-chain.
+
+Participants are able to modify the total balance: the following two functionalities are available:
+
+* deposit - when a participant wants to commit more tokens from his on-chain
+  balance to the channel total balance
+* withdrawal - when a participant wants to transfer tokens out of the channel
+  on-chain balace to one's personal account
```
uwiger: balace -> balance
Should we mention that a deposit/withdrawal of 0 could work as a mutually agreed snapshot operation?

epoch/api/channels_api_usage.md
```diff
@@ -518,3 +521,268 @@ details:
...
+  | state_hash | string | the root of the internal channel state hash after the withdraw |
+  | round | integer | the next channel round |
+  | nonce | integer | withdrawer's nonce |
+
+#### Start withdraw
+
+Any of the participants can initiate a withdrawal. The only requirements are:
+* Channel is already opened
+* No off-chain update/deposit/withdrawal is currently being performed
+* Channel is not being closed or in a solo closing state
```
uwiger: * The withdrawal amount must be equal to or greater than zero, and cannot exceed the available balance on the channel (minus the `channel_reserve`).

epoch/api/channels_api_usage.md
```diff
@@ -518,3 +521,268 @@ details:
...
+  | state_hash | string | the root of the internal channel state hash after the deposit |
+  | round | integer | the next channel round |
+  | nonce | integer | depositor's nonce |
+
+#### Start deposit
+
+Any of the participants can initiate a deposit. Only requirements are:
+* Channel is already opened
+* No off-chain update/deposit/withdrawal is currently being performed
+* Channel is not being closed or in a solo closing state
```
uwiger: * The deposit must be equal to or greater than zero, and cannot exceed the available balance on the depositor's account.

epoch/api/channels_api_usage.md
velzevur: I plan on adding this when describing the solo-snapshot tx

PR #135, opened 2018-06-28
ng intro

sennui:


consensus/consensus.md
```diff
@@ -25,25 +25,37 @@ of all `0`.
...
 the time a block was received at, i.e. it will prefer the block it received first.
 
+
 ### Blocks
 
-A block is a collection of transactions and associated metadata. Blocks are generated by
-miners and new coins can only be created via these new blocks.
+This document depends on Bitcoin-NG protocol that describes rules for building blockchain.
+Bitcoin NG introduces two types of blocks:
+1. Key Blocks - Key Blocks carry an ID of a leader and Proof-of-Work evidence and associated metadata
```
knarz: s/ID/public key/

consensus/consensus.md
```diff
@@ -25,25 +25,37 @@ of all `0`.
 
 At any point, the `block height` is the distance from the current valid `top` block to the
 genesis block. The genesis block has height zero.
+
 A valid block here means valid according to the consensus rules described below.
+
+We use temporary random leader to build the tree of blocks.
```
knarz: What exactly is this supposed to mean?

consensus/bitcoin-ng.md
```diff
@@ -0,0 +1,87 @@
...
+      Leader 1 broadcasted two Micro Blocks. Leader 2 send out new Key Block holding his key.
+      Leader 2 emitted one Micro Block so far
+
+
+Leaders, who follow the protocol, emit Micro Blocks until they receive Key Block broadcasted by another leader. Micro Blocks are not distinguished by height - they inherit the height of the leader that emitted them.
+
+Latency
+==
+
+On average in Bitcoin blocks are broadcasted every 10 minutes - it is driven by protocol and guarded by target. Aeternity assumed 3 minutes for Key Blocks and 3 seconds time delta between Micro Blocks. 3 minutes resolution of Key Blocks is also guarded by target and Proof-of-Work.
```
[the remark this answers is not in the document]
sennui: I don't see good solution for that. We may consider re-validating the whole generation when new key block arrives. At that time we have rough knowledge about time and we can estimate frequency of microblocks. For ongoing arrivals we can only check timestamps that are arbitrary and time drift of those from the recipient's point of view

consensus/consensus.md
sennui: My intent was to follow the convention with "tree" and underline that there is a leader that is responsible for part of structure (microblocks). "Temporary" and "random" should underline that this is not a permissioned setup

consensus/bitcoin-ng.md
```diff
@@ -0,0 +1,87 @@
...
+          ───│MB(1)│────│  KB-2  ├──│MB(2)│
+             └─────┘    │        │  └─────┘
+                        └────────┘
+
+        Micro Block broadcasted by old leader is no longer forming consensus.
+        New leader may read it and include all transaction in newly broadcasted
+        micro blocks. Most likely, the new leader will broadcast micro blocks
+        as soon as possible. Given that nodes in the network have very similar
+        mempool content MB(1)' and MB(2) should be very similar. Re-signing
+        the delta of transactions missing in Leader 2's node is very cheap operation.
```
hanssv: How cheap is it really? We have a new height, etc. so there is validation necessary?!

consensus/consensus.md
```diff
@@ -132,51 +145,77 @@ GENESIS_VERSION: 2
  -------------- ------
 | prev_hash    | 32   |
  -------------- ------
-| root_hash    | 32   |
- -------------- ------
-| trees        |      |
+| state_hash   | 32   |
```
lucafavatella: These serializations are redundant with serializations.md as amended in https://github.com/aeternity/protocol/pull/92/files so may need fine-tuning after merge.

PR #141, opened 2018-07-11
Refresh Sophia example from epoch master

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/158982822
Please review alongside https://github.com/aeternity/epoch/pull/1334

Please refer to commit message for details.

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -611,66 +611,74 @@ In order of highest to lowest precedence.
 
 ```
 /*
- *  A simple crowd funding example.
- *  Not production code (do not use)!
+ * A simple crowd-funding example
  */
 contract FundMe =
 
-  record state = { contributions : map(address, uint),
-                   total         : uint,
+  record spend_args = { recipient : address,
+                        amount    : int }
+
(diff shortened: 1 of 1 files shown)


contracts/sophia.md
```diff
@@ -611,66 +611,74 @@ In order of highest to lowest precedence.
...
+ * A simple crowd-funding example
  */
 contract FundMe =
 
-  record state = { contributions : map(address, uint),
-                   total         : uint,
+  record spend_args = { recipient : address,
+                        amount    : int }
+
+  record state = { contributions : map(address, int),
```
dadaphl: I thought it should be `type` instead of `record`.

lucafavatella: It is because this PR is for (i.e. post 0.17.0) rather than tag 0.17.0

The protocol repo's master [reflects this change since 0.17.0](https://github.com/aeternity/protocol/compare/epoch-v0.17.0...088fcb33b6903d2435a438cf82d4e05cc5fc9d13#diff-9ddeae469112ac9dd44768c45511c6e9R493).

PR #144, opened 2018-07-13
Revert "Illustrate transaction unpacking and signing"

lucafavatella:
Reverts aeternity/protocol#99

### DIFF
--- a/epoch/api/channels_api_usage.md
+++ b/epoch/api/channels_api_usage.md
@@ -68,8 +68,8 @@ in the chain, and the others are metadata used for the connection itself.
 
   | Name | Value |
   | ---- | ----- |
-  | initiator | ak$ngu1K7uGv9rd9GcEWQkLug6REi6MwWjnFM9WpWYYV9tNz5pQn |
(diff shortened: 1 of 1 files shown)


epoch/api/channels_api_usage.md
```diff
@@ -78,19 +78,14 @@ in the chain, and the others are metadata used for the connection itself.
   | ttl | 1000 |
 
   The `initiator` will be connecting to the `responder` on localhost:1234
-  We will be using following tools
-
-  * [wscat](https://github.com/websockets/wscat)
-  * [unpack](https://github.com/aeternity/aepp-sdk-js/blob/develop/examples/unpack.js)
-  * [sign](https://github.com/aeternity/aepp-sdk-js/blob/develop/examples/sign.js)
```
lucafavatella: It is dangling

mpowaga: I believe these utils could be used in following way.

`npm install -g @aeternity/aepp-sdk`

and then

`aecrypto unpack tx$...`

`aecrypto sign tx$...`

However binaries are not defined in [package.json](https://github.com/aeternity/aepp-sdk-js/blob/develop/package.json). What do you think @e-user?

PR #145, opened 2018-07-15
Add serialization and description for the channel_snapshot_solo_tx

velzevur:
Protocol update in correspondence of aeternity/epoch#1325

### DIFF
--- a/channels/ON-CHAIN.md
+++ b/channels/ON-CHAIN.md
@@ -5,6 +5,7 @@
 - [Updates](#updating-channel-on-chain)
 	- [`channel_deposit`](#channel_deposit)
 	- [`channel_withdraw`](#channel_withdraw)
+	- [`channel_snapshot_solo`](#channel_snapshot_solo)
 - [Closing](#closing-channel-on-chain)
 	- [`channel_close_mutual`](#channel_close_mutual)
 	- [`channel_close_solo`](#channel_close_solo)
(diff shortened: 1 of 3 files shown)


channels/ON-CHAIN.md
```diff
@@ -166,6 +167,34 @@ on-chain transactions that have either `round` or a `payload` (with a `round` in
 
 (***TODO***: should a channel be considered closed if all the tokens are taken from it?)
 
+### `channel_snapshot_solo`
```
lucafavatella: Shall this be somewhere in overview https://github.com/aeternity/protocol/blame/01d13252319c6b2ab97356fe9208336243025a15/channels/README.md#L326 ?

channels/README.md
```diff
@@ -334,6 +334,7 @@ channel_settle |   |     open     | -----+
                |   +--------------+
                |    ^ |    |
                |    | | channel_withdraw/
+               |    | | channel_snapshot_solo/
```
lucafavatella: Reading description, I am not convinced this tx makes channel move to closing.

lucafavatella: ah these are referred to arrow in the left

lucafavatella: This suggests that it can be applied only from channel state open, though I do not see this check in [the code](https://github.com/aeternity/epoch/pull/1325/files#diff-ab5fa577fb91bc576556ed547cbef0a4R109)?

lucafavatella: @velzevur Thanks for pointing out it is checked https://github.com/aeternity/epoch/pull/1325/files#diff-a171f3f67fe181f4f1e61696b31151b9R180

PR #151, opened 2018-07-24
change the way the header hash is constructed for pow

knarz:
This updates the formula used to compute the header hash. It was no longer accurate after the update to use Bitcoin-NG, since it included transaction hashes, which are not part of key blocks.

In addition it includes some minor cosmetic changes to adjust line lengths.


consensus/consensus.md
```diff
@@ -280,8 +288,8 @@ EDGEBITS: 28
 
 
 ```
-header = version || height || prev_hash || txs_hash || root_hash || target || evidence || nonce || time
-           64         64         256         256          256         64       42*32       64       64
+header = version || height || prev_hash || root_hash || target || evidence || nonce || time
```
lucafavatella: Miner and beneficiary? ([code](https://github.com/aeternity/epoch/blob/f3676d1d44503cbd3c94c5363f4d37cb04f48cc4/apps/aecore/src/aec_conductor.erl#L320))

See also https://github.com/aeternity/protocol/blame/f0f6c18ee27fd1f349f613c3bb3dbac7034f7ec3/serializations.md#L35-L36

PR #152, opened 2018-07-25
Mention user API for contract call tx with Sophia args

lucafavatella:


lucafavatella:
Delivered https://www.pivotaltracker.com/story/show/159299385

PR #153, opened 2018-07-25
document leave/reestablish + complete other docs

uwiger:


channels/OFF-CHAIN.md
```diff
@@ -291,23 +301,96 @@ different amounts, then that might communicate that it wants these instead.)
...
+ ---------------------- ----
+| length               | 16 |
+ ---------------------- ----
+| data                 | N  |
+ ---------------------- ----
+```
+
+The payload (`data`) must be the latest mutually signed offchain state, and
+the clients must verify that they have the corresponding state trees to match
+the state (otherwise, it will not be possible to close the channel later on.)
```
velzevur: I think the text in brackets could be improved a little bit: no matter the state trees, participants can always make a `channel_close_mutual` that only has the closing balances they've agreed upon (no state hash there).
If they are using different state trees (not matching the `state_hash` provided in the off-chain transaction in `data`) they'd be performing the same updates on different trees and will getting different root hashes for those trees. This would make the channel unusable.
My suggestion would be:
(otherwise, it will not be possible to ~close~ use the channel later on.)

channels/OFF-CHAIN.md
```diff
@@ -68,12 +68,23 @@ The following message codes are defined:
 | channel_open     | 1  |
 | channel_accept   | 2  |
 | channel_reestabl | 3  |
-| funding_created  | 4  |
-| funding_signed   | 5  |
-| funding_locked   | 6  |
-| update           | 7  |
-| update_ack       | 8  |
-| update_error     | 9  |
+| channel_reest_ack| 4
```
velzevur: I see codes match the [epoch ones](https://github.com/aeternity/epoch/blob/master/apps/aechannel/src/aesc_codec.hrl#L30). Ok

channels/OFF-CHAIN.md
```diff
@@ -291,23 +301,96 @@ different amounts, then that might communicate that it wants these instead.)
...
+#### Requirements
+
+*Responder*:
+
+- MUST abort if the `chain_id` doesn't match the current chain.
+- MUST abort if the payload does not correspond to the last known mutually
+  signed state.
+- MUST abort if it doesn't have a corresponding state tree (is able to verify
+  proof-of-inclusion) for the provided state.
+
```
velzevur: We could add something about `Responder` aborting if signature check does not match up (the `data` must be co-signed).
Not required to amend though.

channels/OFF-CHAIN.md
```diff
@@ -316,33 +399,48 @@ initiator to send their signature to the responder.
 
 *Responder*:
 
+- MUST abort if the size of the payload does not match `length`
+- MUST abort if the `data` cannot be deserialized into a valid
+  `channel_create_tx` object
 - MUST abort if the signature is invalid for the provided transaction data
+- SHOULD abort if the `round` of the state is
```
velzevur: Probably missing the part about the round being [> 1](https://github.com/aeternity/epoch/blob/master/apps/aechannel/src/aesc_create_tx.erl#L319)

channels/OFF-CHAIN.md
```diff
@@ -291,23 +301,96 @@ different amounts, then that might communicate that it wants these instead.)
   blockchain to stop a cheater
 - `responder_pubkey` MUST be a valid ed25519 pubkey
 
+### `channel_reestabl`
```
lucafavatella: This causes anchor `#channel_reestabl` rather than `#channel_reestablish` in the outline.

channels/OFF-CHAIN.md
```diff
@@ -291,23 +301,96 @@ different amounts, then that might communicate that it wants these instead.)
...
+to client failure or by the participants mutually agreeing to leave.
+
+```
+  name                  size (bytes)
+ ---------------------- ----
+| chain_hash           | 32 |
+ ---------------------- ----
+| channel_id           | 32 |
+ ---------------------- ----
+| length               | 16 |
```
lucafavatella: 16 or [2](https://github.com/aeternity/epoch/blob/8a996b6082e17a70349076534cbd4fc33a61b129/apps/aechannel/src/aesc_codec.erl#L192)?

```
1> bit_size(<<0:2 /unit:8>>)
1> .
16
2> byte_size(<<0:2 /unit:8>>).
2
```

PR #155, opened 2018-07-30
Contracts in a channel description

velzevur:
Corresponding protocol changes for aeternity/epoch#1385

### DIFF
--- a/channels/OFF-CHAIN.md
+++ b/channels/OFF-CHAIN.md
@@ -988,12 +988,13 @@ Contracts are part of the [update](#state-update) mechanism: it is different
 updates that are being used. Serialization of those can be found [here](../serializations.md#channel-off-chain-update).
 
 First one participant initiates an update round containing a [channel create
-contract update](../serializations.md#channel-off-chain-update-create-contract). It contains all the 
-information needed for a contract creation. The other participant acknowledges
+contract update](../serializations.md#channel-off-chain-update-create-contract). It contains all the
+information needed for a contract creation. The other participant co-signs
 the changes and the contract is considered to be created.
 
 After a contract is created it can be called. For this one of the participants
 initiates an update round containing a [channel call
-contract update](../serializations.md#channel-off-chain-update-call-contract). It contains all the 
-information needed for a contract call, including the contract address. The other participant acknowledges the changes and the contract call is considered to be executed. Its results can be extracted from the calls tree in the state tree.
+contract update](../serializations.md#channel-off-chain-update-call-contract). It contains all the
+information needed for a contract call, including the contract address. The other participant co-signs
+the changes and the contract call is considered to be executed. Its results can be extracted from the calls tree in the state tree.
 
--- a/channels/OFF-CHAIN.md
+++ b/channels/OFF-CHAIN.md
@@ -847,12 +847,6 @@ transaction onto the chain, and then terminate.
  ---------------------- ----
 ```
 
-
-## Initialise contract
-
-### `update_init_contract`
-
-
 ## To be reviewed
 
 A channel can be closed under three circumstances:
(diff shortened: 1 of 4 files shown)


serializations.md
```diff
@@ -532,18 +537,93 @@ The payload is a serialized signed channel offchain transaction or it is empty.
...
-The payload is a serialized signed channel offchain transaction and can not be empty.
+The payload is a serialized signed channel off-chain transaction and can not be empty.
 
-#### Channel offchain transaction
 
-The channel offchain transaction is not included directly in the transaction tree but indirectly as payload of:
+#### Channel off-chain update
+
+Channel update rounds are described by various updates. They all form the
+`update()` type. If not specified something different, any of the addreses
```
lucafavatella: Where is `update()` type specified?

velzevur: This section exactly: all of those form the `update()` type: it could be either transfer, deposit, withdrawal, contract create or contract call.

lucafavatella: Alternative proposal:

> Channel update rounds are described by various updates, that are defined in this subsection. Each mention of type `update()` in the rest of this document is meant to be understood as referring to any one of these updates. ...

channels/OFF-CHAIN.md
```diff
@@ -988,4 +982,18 @@ execute the contract given the state and inputs to produce a new state. This new
...
 
-### Initialisation
+### Contracts lifecycle
+
+Contracts are part of the [update](#state-update) mechanism: it is different
+updates that are being used. Serialization of those can be found [here](../serializations.md#channel-off-chain-update).
+
+First one participant initiates an update round containing a [channel create
+contract update](../serializations.md#channel-off-chain-update-create-contract). It contains all the 
+information needed for a contract creation. The other participant acknowledges
```
knarz: » acknowledges« seems too weak when you mean signing

channels/README.md
```diff
@@ -347,6 +347,25 @@ channel_settle |   |     open     | -----+
...
+Contract life cycle closesly represents the one on-chain: once a contract is
+created - it can be called by participants. Contract has state, balance and a
+list of calles. Participants can make calls and thus modify channel balance
+and state. Participants can inspect the return values of the past calls.
+Participants can remove a contract from a channel's state tree and
+redistribute its balance amongst them.
+
+If a dispute arises - a proof of inclusion of the contract inside channel's
+state tree is posted on-chain in order for the blockchain to enforce progress
+of the channel. That way we keep contracts in channels trustless.
```
knarz: This paragraph could need some more details. How is progress enforced?

velzevur: That would be a subject to another PR - the one that delivers forcing progress on-chain

epoch/api/channels_api_usage.md
```diff
@@ -339,15 +342,246 @@ it and then post it back via a WebSocket message:
...
+      'gas_price': 0,
+			'gas_used': 524,
+      'height': 8,
+      'return_type': 'ok',
+      'return_value': '0x000000000000000000000000000000000000000000000000000000000000002A'
+		}
+}
+```
+
+It is worth mentioning that since this is an off-chain transaction the gas price specified is not consumed.
```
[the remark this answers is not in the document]
velzevur: I've added the following just bellow this line:
>That amount of gas represents the amount of computations. It could be used for
aproximation for the gas needed for executing a contract on-chain if a similar amount of
computations are required. Computation heavy contracts might be just too
expensive to be force progressed on-chain, so please use with caution.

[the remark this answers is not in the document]
knarz: Ideally that should go into the protocol section and not just the usage but that can be changed later

channels/OFF-CHAIN.md
```diff
@@ -988,4 +982,18 @@ execute the contract given the state and inputs to produce a new state. This new
 state could then either be used for both parties to continue operation from there
 or leave it at that.
 
-### Initialisation
+### Contracts lifecycle
```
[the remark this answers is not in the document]
velzevur: Addressed

PR #159, opened 2018-08-09
PT-159590398 Remove support for oracle WS API

jur0:
[PT-159590398](https://www.pivotaltracker.com/n/projects/2124891/stories/159590398)


epoch/api/oracle_api_usage.md
```diff
@@ -13,123 +11,114 @@ to split the example into one node running the oracle and other nodes querying
...
-< {"action":"mined_block","origin":"chain","payload":{"height":1,"hash":"bh$jXjgHkcuXnTY4PtMpctcwFT2jf4fZ1jGdeax1geWoW64hSXpY"}}
-```
+### Register
 
 Once the account has a positive balance we can post an _"Oracle register transaction"_:
 ```
-...
-> {"target":"oracle", "action":"register", "payload":{"type":"OracleRegisterTxObject", "vsn":1, "account":"ak$jzZyCLFtHVD7yVdEhGJFM3LjeXrKqWxnHbCYzhnrrR4DkdF", "query_format":"the query spec", "response_format":"the response spec", "query_fee":4, "oracle_ttl":{"type":"delta", "value":50}, "fee":5, "ttl":1234}}
-< {"action":"register","origin":"oracle","tag":"untagged","payload":{"result":"ok","oracle_id":"ok$jzZyCLFtHVD7yVdEhGJFM3LjeXrKqWxnHbCYzhnrrR4DkdF","tx_hash":"th$26iUqaRt4s1ydAF8z7WDeM4FhCqwEu5TbWTCazYFsPY8Le8Upq"}}
+curl -X POST -H "Content-Type: application/json" http://localhost:3113/v2/oracle-register-tx -d '{"type":"OracleRegisterTxObject", "vsn":1, "account":"ak$EmJyR97vW4jzdcCPCvgjUa8RUmo45E1KnExBum38yz48Frwov", "query_format":"the query spec", "response_format":"the response spec", "query_fee":4, "oracle_ttl":{"type":"delta", "value":50}, "fee":5, "ttl":1234}'
```
lucafavatella: I would suggest refraining from documenting [obsolete paths](https://github.com/aeternity/epoch/blob/dd0da343cd0aa9ab8f4fc23a001430b94775c1ce/config/swagger.yaml#L1760).

lucafavatella:
Thanks for taking care of this!

epoch/api/oracle_api_usage.md
jur0: I used obsolete paths so the example doesn't change too much. Also,  the obsolete paths are used in other docs, too - [spend_api_usage](https://github.com/aeternity/protocol/blob/master/epoch/api/spend_api_usage.md), [naming_system_api_usage](https://github.com/aeternity/protocol/blob/master/epoch/api/naming_system_api_usage.md), etc. These paths are used for posting a tx which is signed by node's key, then included in the chain.

The `/tx/oracle/*` endpoints return an unsigned tx, the client needs to sign it and post using `/tx` endpoint. I haven't investigated how to do signing on the client side.

Anyways, the obsolete endpoints are going to be removed soon and the `tx/oracle/*` are going to be renamed to `/debug/oracles/*` (and eventually removed too as these txs will be created client side only).

PR #161, opened 2018-08-10
adds a stratum specification

knarz:
This adds a specification for a stratum protocol, that allows mining pool operators to delegate work among a group of connected workers.

### DIFF
--- a/STRATUM.md
+++ b/STRATUM.md
@@ -1,0 +1,366 @@
+# Stratum
+
+This document describes a protocol, that allows a group of miners to connect
+to a server, which coordinates the distribution of work packages among
+miners.
+
+The initial protocol was written for Bitcoin and contains several pieces that
+need adjustment in order to be usable with Aeternity.
+
+Pool operators looking to implement this specification should note, that the
+usage of the Bitcoin-NG protocol changes the dynamics of the mining game.
+With Bitcoin-NG pools will end up authoring the microblocks containing
+transactions while the miners try to find keyblocks, which are used for
+leader election.
+This procedure puts both more power into the hands of pool operators while
+also adding more burden. This burden comes in the form of increased book
+keeping complexity—the full revenue of one epoch can only be computed after
+the fact—and an increase in computational resources required for signing
+microblocks.
+
+
+## Specification
+
+
+### Conventions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+The stratum protocol mostly adheres to the [JSON RPC 2.0](https://www.jsonrpc.org/specification) specification.
+
+This specification derives from works by [slushpool](https://slushpool.com/help/manual/stratum-protocol),
+[zcash](https://github.com/str4d/zips/blob/77-zip-stratum/drafts/str4d-stratum/draft1.rst)
+and [nicehash](https://github.com/nicehash/Specifications/blob/master/EthereumStratum_NiceHash_v1.0.0.txt).
+
+
+### Overview
+
+Communication happens over a bidirectional channel with messages encoded as
+JSON with `LF` delimiter—in this document written as `\n`.
+
+Since requests can be handled out of order, each request in a session SHOULD
+have an unique `id`. In order to be able to match request and response,
+responses MUST have the same `id` as their matching request. Notifications
+sent by the server or calls that do not trigger any response MAY have an `id` of
+`null`.
+
+For further details on the members of request and response objects consult the
+[JSON RPC 2.0 specification](https://www.jsonrpc.org/specification).
+
+
+### Protocol flow example
+
+The following shows what a session might look like from subscription to
+submitting a solution.
+
+```
+Client                                Server
+  |                                     |
+  | --------- mining.subscribe -------> |
+  | --------- mining.authorize -------> |
+  | <-------- mining.set_target ------- |
+  |                                     |----
+  | <---------- mining.notify --------- |<--/
+  |                                     |
+  | ---------- mining.submit ---------> |
+```
+
+
+### Methods
+
+- [mining.configure](#mining-configure)
+- [mining.subscribe](#mining-subscribe)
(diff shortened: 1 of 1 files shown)


knarz:
This spec itself is just for general information. Ideally an open source implementation of this is available before mainnet  launch.

sammy007:
Even dummy stratum for mining testing is enough IMO. The standalone CPU miner in C and GPU miner is what needed the most. Without it AE will end up with commercial (dev fee enabled) mining software which is "vulnerable" to pool white listing. I am going to write a pool for it eventually, but currently I have no idea if I will have enough time for that.

It also could be an erlang app (which you can disable) with a simple web UI just like this https://github.com/sammy007/monero-stratum and run along with epoch so mid range and big farms will use it.

STRATUM.md
```diff
@@ -0,0 +1,366 @@
...
+  parameters
+	1. Worker name, mostly for statistics
+	2. Job ID
+	3. Miner nonce in hex format
+	4. Proof of work puzzle solution
+
+As with `mining.notify` all parameters after the job ID should be understood as
+being conditional on the block version sent along in `mining.notify`.
+
+The miner nonce MUST be `8 - len(extra_nonce)` bytes long and be little endian.
```
hanssv: I got confused here, the extranonce was set by the server, why have the length of this field being based on the extra_nonce?

knarz: The miner and extra nonce are concatenated to the full nonce. So if the server sets the extra nonce to be 3 bytes long then the miner must submit 5 bytes nonce with this submit call.

sammy007: Is it ok to have just 8 bytes for nonce? For FPGA for example or for potential ASICs? Usually pools will lease at least 2 bytes (likely 4 bytes) for extra nonce. So only 4 bytes left for miner which could be not enough if miner is very fast (10 GPU set for example).

sammy007: My proposal is to set it to 32 bytes while it's not too late.

knarz: The nonces in our block headers are currently 8 bytes, since there's really no need for a bigger nonce with cuckoo cycle—at least given our current knowledge of available hardware—,although the algorithm has space for the nonces to be bigger.

If we consider a bad case of not producing a block for double the normal block time, i.e. 6 minutes, and a solution taking about 100ms to find, you'd be able to generate about 3600 solutions with a single GPU. At 3600 solutions you'd need to run about 1.2 million GPUs to exhaust those 4bytes.

But, there's a case to be made that future PoW changes might require us to make the nonce space bigger.

sammy007: My point is that 4 bytes will be taken by pool leaving only 4 bytes for miner. Why not just extend it once and forever, extra few bytes in a block is not a problem, a HF to meet hw requirements is. If AE will be a successful project we will see ASIC in 1-2 years, current ASIC performance over CPU on equihash for example is dramatic, so it probably wise to just use 32 bytes since the start and don't bother any more, no need for pools to upgrade consensus in the future, no need for HF, etc.

sammy007: Is there any epoch RPC endpoints responsible for block templates?

STRATUM.md
```diff
@@ -0,0 +1,368 @@
...
+  |                                     |----
+  | <---------- mining.notify --------- |<--/
+  |                                     |
+  | ---------- mining.submit ---------> |
+```
+
+
+### Methods
+
+- [mining.configure](#mining-configure)
```
lucafavatella:
```suggestion
- [mining.configure](#miningconfigure)
```

STRATUM.md
```diff
@@ -0,0 +1,368 @@
...
+- [mining.subscribe](#mining-subscribe)
```
lucafavatella:
```suggestion
- [mining.subscribe](#miningsubscribe)
```

STRATUM.md
```diff
@@ -0,0 +1,368 @@
...
+- [mining.authorize](#mining-authorize)
```
lucafavatella:
```suggestion
- [mining.authorize](#miningauthorize)
```

STRATUM.md
```diff
@@ -0,0 +1,368 @@
...
+- [mining.notify](#mining-notify)
```
lucafavatella:
```suggestion
- [mining.notify](#miningnotify)
```

STRATUM.md
```diff
@@ -0,0 +1,368 @@
...
+- [mining.set_target](#mining-set_target)
```
lucafavatella:
```suggestion
- [mining.set_target](#miningset_target)
```

STRATUM.md
```diff
@@ -0,0 +1,368 @@
...
+- [mining.submit](#mining-submit)
```
lucafavatella:
```suggestion
- [mining.submit](#miningsubmit)
```

STRATUM.md
```diff
@@ -0,0 +1,368 @@
...
+- [client.reconnect](#client-reconnect)
```
lucafavatella:
```suggestion
- [client.reconnect](#clientreconnect)
```

STRATUM.md
```diff
@@ -0,0 +1,368 @@
...
+```
+{ "id": 1, "result": ["ed70df83","080000"], "error": null }\n
+```
+
+- [ `id` : `int` ]: request id
+- [ `result` : (`string`, `string`) ]:
+	- MUST be `null` if an error occurred or otherwise
+		1. If the server supports session resumption, then this SHOULD be a unique
+       session id, `null` otherwise
+    2. The first bytes of the 8 byte nonce, hex encoded
```
lucafavatella: A markup fix (for readability).
```suggestion
		2. The first bytes of the 8 byte nonce, hex encoded
```

lucafavatella: Clarification re order of bytes (and markup fix). A couple of lines below, this doc says that this value "is usually referred to as the extranonce" and notes `nonce = minernonce || extranonce` so seems in contrast with this line. I suggest:
```suggestion
		2. The last bytes of the 8 byte nonce, hex encoded
```

@skkw @sennui I considered:
* [Current dev stratum PR](https://github.com/aeternity/aeternity/pull/2531/files#diff-3f2e90c96b318c0fd0c2dcd5c04a4003R742).
* [Lib used in that PR](https://github.com/aeternity/aestratum_lib/blob/96ffd9a36fab0d65640e87000e8466458276b8f3/src/aestratum_nonce.erl#L198-L207).

PR #164, opened 2018-08-13
document leave/reestablish API

uwiger:
See [PT #159405858](https://www.pivotaltracker.com/story/show/159405858).

Note also that `channel_id` has been added in at least some reports (to make it easier to store `channel_id` + `offchain_tx` – the two elements needed to be able to reestablish a channel.


velzevur:
I see those are the protocol changes regarding aeternity/epoch#1431

PR #170, opened 2018-08-21
Update AEVM primops doc following oracle query fee updates

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/159919056

It should have been part of doc update PR #163 alongside code PR https://github.com/aeternity/epoch/pull/1439

### DIFF
--- a/contracts/aevm.md
+++ b/contracts/aevm.md
@@ -23,19 +23,22 @@ Interaction with the first class objects of the Æternity chain (e.g.
 oracles, names, and state channels) is done through calls to a
 built-in contract at address 0.
 
+The value in the call indicates the tokens that the primop can use from the contract account.
+Unused tokens stay on the contract account.
+
 The first argument in the call specifies which primop to call.
 The following arguments are encoded as Sophia data.
 
-| OpNo | Name                |             Arguments | Return value    |
-| ---- | ------------------- | --------------------- | --------------- |
-|    1 | Spend               | `Recipient : address` | `Nil`           |
-|  100 | Oracle register     | `Acct : address , Sign : signature, QFee : int, TTL : int, QType : typerep, RType : typerep` | `Oracle : address` |
-|  101 | Oracle query        | `Oracle : address, Query : ('a,'b) , QTTL : int, RTTL : int` | `query : oracle_query('a, 'b)` |
-|  102 | Oracle respond      | `Oracle : address, Query : address , Sign : signature, R : 'a` | `()` |
-|  103 | Oracle extend       | `Oracle : address, Sign : signature, Fee : integer, TTL : integer` | `()` |
-|  104 | Oracle get answer   | `Oracle : address, Query : address`   | `option('b)` |
-|  105 | Oracle get question | `Oracle : address, Query : address`   | `'a` |
-|  106 | Oracle query fee    | `Oracle : address`    | `int`           |
+| OpNo | Name                |          Value |             Arguments | Return value    |
+| ---- | ------------------- | -------------- | --------------------- | --------------- |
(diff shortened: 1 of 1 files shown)


contracts/aevm.md
```diff
@@ -23,19 +23,22 @@ Interaction with the first class objects of the Æternity chain (e.g.
 oracles, names, and state channels) is done through calls to a
 built-in contract at address 0.
 
+The value in the call indicates the tokens that the primop can use from the contract account.
```
lucafavatella: See tests:
* sophia_oracles_qfee__query_tx_value_below_qfee_does_not_take_from_poor_oracle
* sophia_oracles_qfee__qfee_in_query_above_qfee_in_oracle_does_not_take_from_poor_oracle
* sophia_oracles_qfee__query_tx_value_below_qfee_does_not_take_from_poor_oracle__remote
* sophia_oracles_qfee__remote_contract_query_value_below_qfee_does_not_take_from_poor_oracle__remote

contracts/aevm.md
```diff
@@ -23,19 +23,22 @@ Interaction with the first class objects of the Æternity chain (e.g.
...
+Unused tokens stay on the contract account.
```
lucafavatella: See tests:
* sophia_oracles_qfee__tx_value_above_qfee_in_query_is_awarded_to_oracle
* sophia_oracles_qfee__tx_value_above_qfee_in_query_is_awarded_to_oracle__remote

lucafavatella: The above two lines are equivalent to stating that "the primop would check that tokens used by the primop is =< value".

contracts/aevm.md
```diff
@@ -23,19 +23,22 @@ Interaction with the first class objects of the Æternity chain (e.g.
 oracles, names, and state channels) is done through calls to a
```
lucafavatella: The necessity to call the built-in contract at address zero in order to interact with the primops makes unnecessary a statement that "the primop would check that ... value =< contract-balance.

contracts/aevm.md
```diff
@@ -23,19 +23,22 @@ Interaction with the first class objects of the Æternity chain (e.g.
...
-|  102 | Oracle respond      | `Oracle : address, Query : address , Sign : signature, R : 'a` | `()` |
-|  103 | Oracle extend       | `Oracle : address, Sign : signature, Fee : integer, TTL : integer` | `()` |
-|  104 | Oracle get answer   | `Oracle : address, Query : address`   | `option('b)` |
-|  105 | Oracle get question | `Oracle : address, Query : address`   | `'a` |
-|  106 | Oracle query fee    | `Oracle : address`    | `int`           |
+| OpNo | Name                |          Value |             Arguments | Return value    |
+| ---- | ------------------- | -------------- | --------------------- | --------------- |
+|    1 | Spend               | `Amount : int` | `Recipient : address` | `Nil`           |
+|  100 | Oracle register     | (Unused.)      | `Acct : address, Sign : signature, QFee : int, TTL : int, QType : typerep, RType : typerep` | `Oracle : address` |
+|  101 | Oracle query        | `QFee : int`   | `Oracle : address, Query : ('a,'b), QTTL : int, RTTL : int` | `query : oracle_query('a, 'b)` |
```
UlfNorell: Preexisting problem, but: We're using `address` everywhere else, but `oracle_query('a, 'b)` here (they're the same for the VM). Also in `oracle_query('a, 'b)`, `'a` is the type of the query and `'b` the type of the response, so the `Query` argument should have type `'a`.

lucafavatella: Thanks, will address.

PR #171, opened 2018-08-21
Update consensus.md to describe DigiShield target adjustment

hanssv:


consensus/consensus.md
```diff
@@ -425,12 +425,26 @@ EstimatedCapacity = Sum(EstimatedCapacity[i] * MiningTime[i]) / TotalTime
                   = Sum(HIGHEST_TARGET / Target[i]) / TotalTime
 ```
 
+To get a good trade-off between response time and stability we use the
+[DigiShield v3 algorithm](https://github.com/zawy12/difficulty-algorithms/issues/9)
```
knarz: Maybe note that this is a slightly modified version of the original algorithm

hanssv:
Reformulated DigiShield reference slightly.

consensus/consensus.md
```diff
@@ -134,8 +134,8 @@ nonce:
 #### Key Blocks
 
 ```
-PROTOCOL_VERSION: 21
-GENESIS_VERSION: 21
+PROTOCOL_VERSION: 22
+GENESIS_VERSION: 22
 ```
 
 - the timestamp of a key block MUST be smaller than `now() + 10m`
```
lucafavatella: Two lines below: "the last 11 blocks" means "the last 11 key blocks", right?

consensus/consensus.md
```diff
@@ -400,7 +400,7 @@ DesiredRate              = 1 / DesiredTimeBetweenBlocks
 ```
 
 The basic idea of the algorithm is to estimate the current network capacity
-based on the `N` (= 10) previous blocks and use that to set the new
+based on the `N` (= 17) previous blocks and use that to set the new
```
lucafavatella: This line is implicit on [which block the target adjustment starts at (that seems from block at height 19 - included - the target is adjusted)](https://github.com/aeternity/epoch/pull/1460/files?utf8=%E2%9C%93&diff=split&w=1#diff-4ac652308bacbca71e5f388759d362c1R369). Making it explicit somewhere in the document would remove a little room for ambiguity.

And it seems that idea is excluding blocks 0 (genesis) and 1 from target adjustment?

(This document already details that genesis has height zero - that is good.)

consensus/consensus.md
lucafavatella: T = 3 minutes  # Target solvetime.
FTL = 10 minutes  # Future time limit.

From https://github.com/zawy12/difficulty-algorithms/issues/9#issue-280207016
> … where FTL = future time limit nodes allow blocks to have and is 3xT as described on the [LWMA page](https://github.com/zawy12/difficulty-algorithms/issues/3). Otherwise there is a timestamp exploit by > 50% miners that can greatly lower difficulty.

So recommended FTL is 9 minutes while we use 10.

lucafavatella: I did not find any requirements on adapting the 11 timestamp median-related number of blocks at the line below with the 17 of the target adjustment. So this does not seem requiring change.

hanssv:
Pushed a clarification that the first block we adjust target for according to the algorithm is block 19 `(N + 2)`.

consensus/consensus.md
```diff
@@ -413,24 +413,40 @@ We can estimate the network capacity used to mine a given block `i` as
 
 ```
 EstimatedCapacity[i] = Difficulty[i] / MiningTime[i]
-MiningTime[i]        = Time[i + 1] - Time[i]
+MiningTime[i]        = Time[i] - Time[i - 1]
 ```
 
 The estimated capacity across all `N` blocks is then the weighted (by time)
-average of the estimated capacities for each block.
+average of the estimated capacities for each block. Note, since `MiningTime[i]`
```
lucafavatella: I see patch pushed:

> Note, since `MiningTime[i]`
> require `Time[i - 1]` and we do not use the genesis block timestamp in target
> adjustment we cannot start using the target estimation until block `N + 2 (= 19)`.

lucafavatella: As genesis block has height 0, the first 17 blocks to consider are from 2 to 18 (hence prev timestamps 1 to 17). So first block that shall have target not necessarily equal to predecessor is 19 - correct.

lucafavatella: Current code in verification expects that [block 18 has target copied](https://github.com/aeternity/epoch/blob/0eb614be03ce0cccbbdf7c58420bbdda7b739249/apps/aecore/src/aec_chain_state.erl#L366) that seems correct.

lucafavatella: Current code in generation [if top hash height 18](https://github.com/aeternity/epoch/blob/0eb614be03ce0cccbbdf7c58420bbdda7b739249/apps/aecore/src/aec_block_key_candidate.erl#L55) (hence candidate will have height 19) it triggers target adjustment.

lucafavatella: So code looks correct (19 first block with adjusted target) in both verification and generation.

consensus/consensus.md
knarz: Increasing the FTL would mean that miners have more freedom in manipulating the MTP, so I'd say we stick with the recommended 9 minutes.

lucafavatella: Tracked as https://www.pivotaltracker.com/story/show/160129190

lucafavatella: Approving - tracked this as https://www.pivotaltracker.com/story/show/160129190

PR #174, opened 2018-08-24
PT-160016525 contract types

UlfNorell:
Companion PR for https://github.com/aeternity/epoch/pull/1479

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -238,10 +238,10 @@ The following builtin functions are defined on maps:
 
 #### Account interface
 
-To spend tokens from the contract account to the account "to" you call the raw_spend function.
+To spend tokens from the contract account to the account "to" you call the `Chain.spend` function.
 
 ```
-raw_spend(to : address, amount : integer)
+Chain.spend(to : address, amount : integer)
 ```
 
 #### Oracle interface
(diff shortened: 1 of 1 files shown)


contracts/sophia.md
```diff
@@ -59,6 +59,40 @@ worked out).
...
+    v.vote(alt)
+```
+
+Contract calls take two optional named arguments `gas : int` and `value : int`
+that lets you set a gas limit and provide tokens to a contract call. If omitted
+the defaults are no gas limit and no tokens. Suppose there is a fee for voting:
+
+```javascript
+  function voteTwice(v : VotingType, fee : int, alt : string) =
+    v.vote(value = fee, alt)
```
lucafavatella: If I understand this correctly, before this line I would expect check that `(2 * fee) =< Call.value`. Or a comment mentioning not to copy-paste this code in production?

contracts/sophia.md
```diff
@@ -59,6 +59,40 @@ worked out).
...
+    v.vote(value = fee, alt)
+```
+
+Named arguments can be given in any order.
```
lucafavatella: Does this imply that named arguments cannot be given by position?

contracts/sophia.md
```diff
@@ -453,7 +468,7 @@ The block-chain environment available to a contract is defined in three name spa
   calling the contract.
 - `Call.value` is the amount of coins transferred to the contract in the call.
 - `Call.gas_price` is the gas price of the current call.
-- `Call.gas_left` is the amount of gas left for the current call.
+- `Call.gas_left()` is the amount of gas left for the current call.
 - `Chain.get_balance(a : address)` returns the balance of account `a`.
```
lucafavatella: It feels a bit strange for the reader that `Chain.spend` is in namespace Chain though is not listed outside of this section. Though I do not have better proposal so ok.

PR #176, opened 2018-08-24
Update serialization formats of blocks and headers

gorillainduction:
https://www.pivotaltracker.com/story/show/158659424

### DIFF
--- a/serializations.md
+++ b/serializations.md
@@ -17,11 +17,10 @@ user API.
 
 There are two types of blocks:
 
-* key block - does not contain a list of transactions, therefore, it is
-not divided into a header and a list of transactions;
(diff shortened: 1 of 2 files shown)


serializations.md
```diff
@@ -175,6 +170,8 @@ subsequent sections divided by object.
 | Channel | 58 |
 | Channel snapshot transaction | 59 |
 | POI | 60 |
+| Key block | 100 |
+| Micro block | 101 |
```
hanssv: Would you mind adding `| Light micro block | 102 |` or else we get a conflict there...

gorillainduction: I can do that.

PR #181, opened 2018-08-28
Revisit the closing section

velzevur:
PT [Protocol: revisit channel closing section](https://www.pivotaltracker.com/story/show/159365180)


channels/OFF-CHAIN.md
```diff
@@ -883,8 +886,8 @@ A                       B
 ```
 
 In case of both parties want to close the channel in agreement and
-there is enough money left in the channel to pay for the fee, then the
-the advised distribution of returning the funds left in the channel is
+there are enough tokens left in the channel to pay for the fee, then the
```
knarz: I would say we stick with coins instead of tokens.

velzevur: Replaced tokens with coins everywhere in the file

PR #185, opened 2018-08-30
Fine tune oracle tx fees in oracle usage

lucafavatella:


epoch/api/oracle_api_usage.md
```diff
@@ -79,14 +79,13 @@ We can post a query to the existing oracle:
...
 ```
-**NOTE** Similarly to the oracle registration transaction, the _transaction fee_ depends
-on the _TTL_ **and** the oracle query fee. The base query transaction fee is 2,
-we set the oracle query fee to 4 when registering, and the TTL accounts for 1
-per 1000 blocks (just a preliminary value to test the concept). I.e. the cost
-for our test transaction is 7.
+**NOTE** Similarly to the oracle registration transaction, the _transaction fee_ field depends
+on the _TTL_. The base query transaction fee is 1 (as all other transactions),
+and the TTL accounts for 1 per 1000 blocks (just a preliminary value to test the concept). I.e. the minimum fee field for our test transaction is 2.
+As the oracle was registered with query fee 4 when registering, the code of the transaction is be 6.
```
jur0: `the code of the transaction is be 6.`?

lucafavatella: Addressed in 64377cd

PR #186, opened 2018-08-30
WS API fixes

jur0:
- [x] change block prefixes
- [x] add signature
- [x] add beneficiary
- [x] remove any mentions of msg pack
- [x] fix required "No" in prev_hash and pow in key block


velzevur:
Not sure if this needed, but yet - nice to have

PR #189, opened 2018-09-05
SC: remove push_amount

velzevur:
We no longer have `push_amount` but there had been some hanging data for it.
This PR addresses changes introduced by aeternity/epoch#1526


velzevur:
Since `push_amount` had been fixed in aeternity/epoch#1531, I am closing this PR

PR #190, opened 2018-09-06
Document COINBASE.

happi:
Update protocol to reflect that a COINBASE instruction returns the beneficiary of previous keyblock.


lucafavatella:
I see two approvals. I see https://github.com/aeternity/epoch/pull/1502 merged.

PR #194, opened 2018-09-11
adds fixed rent for name entries

knarz:
This change introduces a simple, fixed rent model to the naming system plus some minor cleaning up.

### DIFF
--- a/AENS.md
+++ b/AENS.md
@@ -236,6 +236,9 @@ pre-claim | |              ||  _
 Note that `expire` is not an explicit message that is part
 of the protocol.
 
+The accurate binary representations of the messages involved can be found in the
+[serializations document](./serializations.md).
+
 
 #### Pre-claim
 
@@ -291,7 +294,8 @@ A `claim` transaction MUST NOT be in included in the same block as its
 `rent` is the reoccurring fee that has to be paid for names. Our first rent
 model is a naïve one, that just uses a flat fee. Our initial target is 4 aeons
 per year. This comes out to `4/floor(60 * 24 * 365/key_block_interval)` aeons
-per block.
+per block. This rent is paid up front.
+
 
 #### Update
 
@@ -299,7 +303,7 @@ per block.
  ------------ ----
 | hash       | 32 |
  ------------ ----
-| expire_by  | 8  |
+| name_ttl   | 8  |
  ------------ ----
 | client_ttl | 8  |
  ------------ ----
@@ -312,7 +316,7 @@ per block.
 The `update` transaction MUST be signed by the owner of the name entry to be
 updated.
 
-The `expire_by` MUST NOT be more than `floor(60 * 24 * (365/2)/key_block_interval)`,
+The `name_ttl` MUST NOT be more than `floor(60 * 24 * (365/2)/key_block_interval)`,
 i.e. half a year, blocks into the future.
 
 `update` transaction may be used to extend the lease of the name.
@@ -320,7 +324,7 @@ i.e. half a year, blocks into the future.
 `rent` is the reoccurring fee that has to be paid for names. Our target is to
 charge 4 aeons per year. This comes out to `4/floor(60 * 24 * 365/key_block_interval)`
 aeons per block. For an update, the rent charged is then
-`(new_expire_by - old_expire_by) * 4/floor(60 * 24 * 365/key_block_interval)`
+`(new_expire_by - old_expire_by) * 4/floor(60 * 24 * 365/key_block_interval)`.
 
 
 #### Transfer
@@ -345,8 +349,7 @@ transferred.
  ------------ ----
 ```
 
-The revoke transaction MUST be signed by the owner
-of a name entry.
+The revoke transaction MUST be signed by the owner of a name entry.
 
 After the `revoke` transaction has been included in the chain, the name enters
 the `revoked` state. After a fixed timeout of `floor(60 * 24 * 14/key_block_interval)`
--- a/AENS.md
+++ b/AENS.md
@@ -52,9 +52,9 @@ the general public can start claiming names.
 
 ## Mechanisms
 
-It is unclear what a good mechanism for a naming system would look
-like. If we imagine two actors both being interested in the same name,
-what would a »fair« solution be to resolve this?
+It is unclear what a good distribution mechanism for a naming system would look
+like. If we imagine two actors both being interested in the same name, what
+would a »fair« solution be to resolve this?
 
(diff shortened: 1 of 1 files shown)


AENS.md
```diff
@@ -252,8 +252,8 @@ A client interacting with the blockchain should generate a warning
 if a user tries to claim a name that is not available.
 
 The `pre-claim` has an implicit expiration attached to it. A
-`pre-claim` MUST be considered invalid after 300 blocks, i.e.
-about two days at 10 minute block time.
+`pre-claim` MUST be considered invalid after `floor(60 * 24 * 2/key_block_interval)` blocks, i.e.
```
lucafavatella: I see this may require change at https://github.com/aeternity/epoch/blob/e9e3b000475b58501fc0daf8e8b1d57ea8b25669/apps/aecore/src/aec_governance.erl#L101

AENS.md
```diff
@@ -268,24 +268,30 @@ commitment := Hash(NameHash(name) + name_salt)
...
 
-The `claim` transaction MUST be signed by the same private key as a
-`pre-claim` transaction containing a commitment to the name and nonce.
+The `claim` transaction MUST be signed by the same private key as a `pre-claim`
+transaction containing a commitment to the name and nonce.
 
 A `claim` transaction MUST NOT be in included in the same block as its
 `pre-claim`.
 
+`rent` is the reoccurring fee that has to be paid for names. Our first rent
```
lucafavatella: Is this rent the total amount? Or the amount per (key)block?

lucafavatella: Do I understand it correctly that the total amount (i.e. considering all blocks the name is rented for) needs to be charged all at time of inclusion of this claim transaction?

lucafavatella: For how long is the name meant to be rented for? And is the user meant to be able to control that when claiming?

I see [code currently assumes relative TTL of 50000](https://github.com/aeternity/epoch/blob/e9e3b000475b58501fc0daf8e8b1d57ea8b25669/apps/aens/src/aens_claim_tx.erl#L140-L142) though not sure whether this is specified anywhere.

AENS.md
```diff
@@ -297,18 +303,25 @@ A `claim` transaction MUST NOT be in included in the same block as its
...
 ```
 
-The `update` transaction MUST be signed by the owner
-of the name entry to be updated.
+The `update` transaction MUST be signed by the owner of the name entry to be
+updated.
 
-The `expire_by` MUST NOT be more than 36000 blocks into
-the future.
+The `expire_by` MUST NOT be more than `floor(60 * 24 * (365/2)/key_block_interval)`,
```
lucafavatella: The normative serialization document serializations.md (in this repo) calls this relative TTL `name_ttl`. It may help the reader to link in this document the serializations in serializations.md and use here the same field names as in serializations.md when describing them - as here.

lucafavatella: The [code limits this relative TTL to 50000 blocks](https://github.com/aeternity/epoch/blob/e9e3b000475b58501fc0daf8e8b1d57ea8b25669/apps/aens/src/aens_update_tx.erl#L241-L242). Is this specified anywhere?

AENS.md
```diff
@@ -297,18 +303,25 @@ A `claim` transaction MUST NOT be in included in the same block as its
...
+i.e. half a year, blocks into the future.
 
 `update` transaction may be used to extend the lease of the name.
-We do not require an additional fee for extending the lease.
+
+`rent` is the reoccurring fee that has to be paid for names. Our target is to
```
lucafavatella: Do I understand it correctly that the total amount (i.e. considering all blocks the name is rented for) needs to be charged all at time of inclusion of this update transaction?

AENS.md
```diff
@@ -335,17 +348,17 @@ of the name entry to be transfered.
 The revoke transaction MUST be signed by the owner
 of a name entry.
 
-After the `revoke` transaction has been included in the chain,
-the name enters the `revoked` state. After a fixed timeout of
-2016 blocks, the name will be available for claiming again.
+After the `revoke` transaction has been included in the chain, the name enters
+the `revoked` state. After a fixed timeout of `floor(60 * 24 * 14/key_block_interval)`
```
lucafavatella: I see the code may need to change from the [current 2016 in revoke transaction](https://github.com/aeternity/epoch/blob/e9e3b000475b58501fc0daf8e8b1d57ea8b25669/apps/aens/src/aens_revoke_tx.erl#L114-L116).

AENS.md
```diff
@@ -335,17 +348,17 @@ of the name entry to be transfered.
...
+blocks, the name will be available for claiming again.
```
lucafavatella: I see from the code that there is [same 2016 timeout when name expires](https://github.com/aeternity/epoch/blob/e9e3b000475b58501fc0daf8e8b1d57ea8b25669/apps/aens/src/aens_state_tree.erl#L223-L225). This may need to be clarified in this specification, as it seems not clear (see also where `revoked` state is mentioned in "Mechanisms" section).

AENS.md
hanssv: That is the only time we can know if the transaction creator has funds, so I guess there is no other alternative.

knarz: Rent is charged up front, yes

AENS.md
knarz: That specific limit is not specified anywhere.

knarz: I'm actually not sure why it is even called `name_ttl` here and then `expires` in the tree. Seems inconsistent.

AENS.md
```diff
@@ -268,47 +271,61 @@ commitment := Hash(NameHash(name) + name_salt)
...
-`pre-claim` transaction containing a commitment to the name and nonce.
+The `claim` transaction MUST be signed by the same private key as a `pre-claim`
+transaction containing a commitment to the name and nonce.
 
 A `claim` transaction MUST NOT be in included in the same block as its
 `pre-claim`.
 
+`rent` is the reoccurring fee that has to be paid for names. Our first rent
+model is a naïve one, that just uses a flat fee. Our initial target is 4 aeons
+per year. This comes out to `4/floor(60 * 24 * 365/key_block_interval)` aeons
```
lucafavatella: Assuming [transaction fees as gas](https://www.pivotaltracker.com/story/show/160256699), rent here will need to be expressed as gas.

zp-sd:
I guess we could flag this with some `post-mainnet` tag?

ThomasArts:
This has been open for a year and names have changed. I propose to close it and revisit the actual documentation in the rewrite that's ongoing. @sennui

PR #198, opened 2018-09-12
Clarify order of record values in memory layout

lucafavatella:
Please refer to commit message for details.

Comes from code review at https://github.com/aeternity/epoch/pull/1494/files#r216999564

PR #199, opened 2018-09-12
Review of readability of Sophia memory layout

lucafavatella:
Prepared while preparing #198.

TODO:
* [ ] Once #198 merged, edit base branch to master.

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -880,9 +880,9 @@ Data is encoded in memory as follows:
   - Records are encoded as a tuple of field values.
     - The order of values is specified as the order the fields appear in the record definition.
   - Maps are encoded as a list of key-value pairs. ***This is subject to change***
-  - The order of values is unspecified. For instance, in the encoding of a pair
-    of boxed values, the three cells (first component, second component, and
-    pair cell) can appear in any order in the encoded binary.
+    - The order of values is unspecified. For instance, in the encoding of a pair
+      of boxed values, the three cells (first component, second component, and
+      pair cell) can appear in any order in the encoded binary.
   - The option type is encoded as -1 for `None` and as a singleton tuple `{v}` for `Some(v)`.
   - Values of datatypes are encoded as tuples where the first component is a
     constructor tag (starting with 0 for the first constructor), and the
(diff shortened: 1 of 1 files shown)


contracts/sophia.md
```diff
@@ -876,13 +876,13 @@ Data is encoded in memory as follows:
...
   - Lists are encoded by an unboxed -1 word for the empty list and an encoded
-    pair of the head and the tail for a cons cell.
+    pair (i.e. tuple of two components) of the head and the tail for a cons cell.
   - Records are encoded as a tuple of field values.
     - The order of values is specified as the order the fields appear in the record definition.
   - Maps are encoded as a list of key-value pairs. ***This is subject to change***
-  - The order of values is unspecified. For instance, in the encoding of a pair
-    of boxed values, the three cells (first component, second component, and
-    pair cell) can appear in any order in the encoded binary.
+    - The order of values is unspecified. For instance, in the encoding of a pair
```
lucafavatella: I am not confident about the content though it seems to be subpoint of maps point.

happi: Perhaps it should read:
"The position of a value in the binary is unspecified."
I think what we are trying to convey is that:
The tuple (17, 42) might be laid out in memory so that the addresses of components is unordered e.g. the tuple might be at address 10 and the content at address 8 and 13:
address: value
  8: 17
  9: xx
10: 8
11: 13
12: xx
13: 17

lucafavatella: @happi Thanks, your comment helped me in reading it better.

> The tuple (17, 42) ...

I understand that the binary that encodes a tuple with two components has to have the two components - either unboxed values or pointers - "stored left-to-right" *adjacently*. If the two components are boxed, the two pointers are adjacent while the order of the three pointers (i.e. location of the tuple, location of the first component, location of the second component) can be whatever.

17 and 42 are unboxed so I expect they must be adjacent (and in order) in memory (because part of a tuple with two components).

lucafavatella: I will update this branch in the following days.

UlfNorell: This part has been completely rewritten in https://github.com/aeternity/protocol/pull/207.

lucafavatella:
I see rewritten and clarified in master (03caba50223b8a163a3ff60576e27bbf0f14a993). Closing.

PR #201, opened 2018-09-18
Pt 155700135 proof of fraud

gorillainduction:
https://www.pivotaltracker.com/story/show/155700135
https://github.com/aeternity/epoch/pull/1557

### DIFF
--- a/consensus/bitcoin-ng.md
+++ b/consensus/bitcoin-ng.md
@@ -96,9 +96,13 @@ To prevent this malicious behaviour, the miner of the next key block
 can submit Proof-of-Fraud in a micro block in its generation. Only
 this leader can submit Proof-of-Fraud to avoid flooding the network
 with reports, which in itself could be viewed as a DoS
-attack. Furthermore, reporting a key block more than once is regarded
-as a protocol violation, even if there is more than one act of fraud
-committed.
+attack.
+
+Furthermore, reporting fraud for a given generation more than once is
+regarded as a protocol violation, even if there is more than one act
+of fraud committed. Under the constraint that only blocks from the
+previous generation can be reported, this means that only one fraud
+report is allowed in a generation.
 
 The Proof-of-Fraud is constructed of two sibling micro headers (with
 the same previous pointer) along with the public key of the
@@ -109,8 +113,7 @@ existence in the block.
 If a miner is reported as fraudulent in this manner, its beneficiary
 will not get any mining reward for that key block (i.e., no coinbase
 and no fees for either generation). The reporter of the fraud will get
-a reward controlled by governance (initially set to 5% of the coinbase
-reward).
+a reward controlled by governance.
 
 The serialization format of the Proof-of-Fraud object can be found
 [here](../../serializations.md#proof-of-fraud), and the inclusion of
--- a/consensus/bitcoin-ng.md
+++ b/consensus/bitcoin-ng.md
@@ -84,9 +84,41 @@ Bitcoin-NG inherits Bitcoin's ability to fork in case of multiple block producer
         mempool content MB(1)' and MB(2) should be very similar. Re-signing
         the delta of transactions missing in Leader 2's node is very cheap operation.
 
+Proof of Fraud
+==
 
+A malicious leader could construct forks in a generation of micro
+blocks, either by forking directly from his key block, or by forking
+on a micro block. This could be done to disrupt the network or to
+perform double spend attacks.
 
-References:
+To prevent this malicious behaviour, the miner of the next key block
+can submit Proof-of-Fraud in a micro block in its generation. Only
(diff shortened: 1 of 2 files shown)


serializations.md
```diff
@@ -722,7 +741,19 @@ NOTE: As the POI contains the Merkle Patricia Tree nodes (e.g. not only their ha
 
 #### Micro Body
 ```
-[ <transactions> :: [binary()]
+[ <transactions>    :: [binary()]
+, <proof_of_fraud>> :: [binary()]
```
lucafavatella: I see this section is referred to by MSG_MICRO_BLOCK P2P message so that P2P message description seems up to date. Ok.

consensus/bitcoin-ng.md
```diff
@@ -84,9 +84,41 @@ Bitcoin-NG inherits Bitcoin's ability to fork in case of multiple block producer
...
+A malicious leader could construct forks in a generation of micro
+blocks, either by forking directly from his key block, or by forking
+on a micro block. This could be done to disrupt the network or to
+perform double spend attacks.
+
+To prevent this malicious behaviour, the miner of the next key block
+can submit Proof-of-Fraud in a micro block in its generation. Only
+this leader can submit Proof-of-Fraud to avoid flooding the network
+with reports, which in itself could be viewed as a DoS
+attack. Furthermore, reporting a key block more than once is regarded
```
lucafavatella: "a key block" reads not well for some reasons. Maybe

> reporting a key block more than once

shall be replaced by

> reporting more than a fraud in a generation

?

lucafavatella: Even "reporting more than a fraud in a generation" is a bit ambiguous - as it does not clarify whether it is the fraud or the report that cannot be more than one in a generation.

consensus/bitcoin-ng.md
```diff
@@ -84,9 +84,41 @@ Bitcoin-NG inherits Bitcoin's ability to fork in case of multiple block producer
...
+as a protocol violation, even if there is more than one act of fraud
+committed.
+
+The Proof-of-Fraud is constructed of two sibling micro headers (with
+the same previous pointer) along with the public key of the
+signer. The Proof-of-Fraud is included in a micro block body, and the
+header must contain the hash of the Proof-of-Fraud object to mark its
+existence in the block.
```
lucafavatella: So there is no constraint that at least one of the two reported microblock has been included in the chain. That makes sense. Ok.

consensus/bitcoin-ng.md
```diff
@@ -84,9 +84,41 @@ Bitcoin-NG inherits Bitcoin's ability to fork in case of multiple block producer
...
+
+If a miner is reported as fraudulent in this manner, its beneficiary
+will not get any mining reward for that key block (i.e., no coinbase
+and no fees for either generation). The reporter of the fraud will get
+a reward controlled by governance (initially set to 5% of the coinbase
```
lucafavatella: Shall the 5% be not the 5% of the coinbase only, but the 5% of the full compensation being invalidated?


From https://www.usenix.org/system/files/conference/nsdi16/nsdi16-paper-eyal.pdf

> The poison transaction has to be placed on the blockchain within the maturity window of the misbehaving leader’s key block, and before the revenue is spent by the malicious leader. Besides invalidating the compensation sent to the leader that generated the fork, a poison transaction grants the current leader a fraction of that compensation, e.g., 5%. The choice of this value is explained in Section 5.

The whitepaper does not seem to actually explain this 5% value in section 5. Though I understand that the 5% shall be 5% of "the compensation sent to the leader that generated the fork" that is getting invalidated ("Besides invalidating the compensation ... a fraction of *that* compensation ..."). Linking the fraud report reward only to the coinbase would link the economic incentive of the PoF reported only to the inflation - if any.

lucafavatella: Subtle point, please consider this as optional. The sentence "controlled by governance ... initially set to 5% of the coinbase reward" keeps ambiguity on whether this is meant to be the coinbase from governance at the height of the PoF report, or the coinbase that was meant to be rewarded (that - unless of hard fork stating otherwise - I expect shall be the coinbase from governance at the height of the mined key block). (This comments leaves fees out - for those please refer to [other comment](https://github.com/aeternity/protocol/pull/201/files#r218380643).)

gorillainduction: Valid point. This needs to be revisited. I suggest accepting this as it stands, since it reflects the implementation. We can have a follow up ticket where this is changed if we come to the conclusion that the 5% should indeed be of the total reward. Both for code and documentation.

lucafavatella: Same as https://github.com/aeternity/epoch/pull/1557#discussion_r218406880
Ok.

PR #206, opened 2018-09-21
 PT-160129190 Change accepted future block time shift from 10 to 9 min

jur0:
[PT-160129190](https://www.pivotaltracker.com/n/projects/2124891/stories/160129190)

### DIFF
--- a/consensus/consensus.md
+++ b/consensus/consensus.md
@@ -136,7 +136,7 @@ PROTOCOL_VERSION: 22
 GENESIS_VERSION: 22
 ```
 
-- the timestamp of a key block MUST be smaller than `now() + 10m`
+- the timestamp of a key block MUST be smaller than `now() + 9m`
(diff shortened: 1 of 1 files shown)


consensus/consensus.md
```diff
@@ -374,7 +374,7 @@ TemperedTotalTime = 0.75 * N * DesiredTimeBetweenBlocks + 0.2523 * TotalTime
 ```
 
 Where FTL = Future Time Limit - i.e. the time a block is allowed to be
-"from the future". We use 10 minutes (600 s).
+"from the future". We use 9 minutes (540 s).
```
lucafavatella: A little note like "We use 9 minutes (540 s) i.e. 3 times the DesiredTimeBetweenBlocks." would help the reader though optional.

jur0: It's updated now.

PR #211, opened 2018-09-26
Describe BLOCKHASH

gorillainduction:
https://www.pivotaltracker.com/story/show/160732539
Code PR: https://github.com/aeternity/epoch/pull/1591

### DIFF
--- a/contracts/aevm.md
+++ b/contracts/aevm.md
@@ -57,4 +57,9 @@ generation stored in the previous key block.
 When executing a contract in a state channel (either off-chain or in
 a force progress) DIFFICULTY will return 0.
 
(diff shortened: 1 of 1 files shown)


contracts/aevm.md
```diff
@@ -57,4 +57,9 @@ generation stored in the previous key block.
 When executing a contract in a state channel (either off-chain or in
 a force progress) DIFFICULTY will return 0.
 
-
+The BLOCKHASH instruction returns the hash of the key block at the specified height.
```
hanssv: It is more or less always 0, occasionally it is the hash of the key block ;-)

PR #213, opened 2018-09-27
Signatures in sophia

hanssv:
To allow a contract to perform Oracle and AENS operations on behalf of a third
    party we need to provide and check a Signature.

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -134,7 +134,7 @@ Sophia has the following types:
 | state        | A record of blockstate key, value pairs  |
 | transactions | An append only list of blockchain transactions |
 | events       | An append only list of blockchain events (or log entries) |
-| signature    | A signature - 64 bytes - represented as two 256-bit words. | ```{81222...22336, 90751...60096}```
+| signature    | A signature - 64 bytes |
 | Chain.ttl    | Time-to-live (fixed height or relative to current block) | ```FixedTTL(1050)``` ```RelativeTTL(50)```
 | oracle('a, 'b)       | And oracle answering questions of type 'a with answers of type 'b |  ```Oracle.register(acct, sign, qfee, ttl)```
 | oracle_query('a, 'b) | A specific oracle query |  ```Oracle.query(o, q, qfee, qttl, rttl)```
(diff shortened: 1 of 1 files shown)


contracts/sophia.md
```diff
@@ -134,7 +134,7 @@ Sophia has the following types:
 | state        | A record of blockstate key, value pairs  |
 | transactions | An append only list of blockchain transactions |
 | events       | An append only list of blockchain events (or log entries) |
-| signature    | A signature. |
+| signature    | A signature - 64 bytes |
```
lucafavatella: Could you please add in the `Example` column how a literal signature would be written in Sophia?

UlfNorell: You can't write a literal signature in Sophia. You have to pass it to the contract as an argument.

lucafavatella: If I understand this correctly, you can write a literal signature in Sophia (explanation below) though you are not meant to. So it looks like this intended usage shall be documented. (I did not find it in this PR.) Not sure if this doc is meant to be in scope of any other tickets.

----

Based on my understanding of https://github.com/aeternity/epoch/pull/1600 (I started looking at it) user in Sophia (of course in bytecode can do whatever) can:
* Define contract A with function F that takes argument of type `signature` and calls primop that requires type `signature` (so type checker is happy).
* Create contract A on chain at address AA.
* Define contract B with - in it - declaration of contract A function F ([example](https://github.com/aeternity/epoch/blob/cbb26b2f8588014efc719bb8f9337f52dba62175/apps/aesophia/test/contracts/remote_oracles.aes#L5) - not sure if documented), declaring the argument of F to be not `signature` but `{uint, uint}` (or `{int, int}`... any word type), and defining function in B that calls A.F at address AA passing literal tuple e.g. `{0, 0}` (or whatever literal) as signature. (Signature check will most likely fail - my point is that user can - technically speaking - write literal signature.)

contracts/sophia.md
```diff
@@ -317,14 +317,16 @@ To register a new oracle answering questions of type `'a` with answers of type `
...
                 ttl  : Chain.ttl) : oracle('a, 'b)
 ```
 
 * The `acct` is the address of the oracle to register (can be the same as the contract).
-* The `sign` is a signature of the address to register proving you have the private key
-  of the oracle, or the integer 0 when address is the same as the contract.
+* The `sign` is a signature proving that the contract is allowed to register the account -
+  the account address + the contract address (concatenated as byte arrays) is signed with the
+  private key of the account, proving you have the private key of the oracle to be. If the
+  address is the same as the contract `sign` is 0.
```
UlfNorell: 0 is not a valid signature. The nicest solution is probably to make the signature a named argument, but for now maybe change to "If the address is the same as the contract the signature is ignored".

contracts/sophia.md
hanssv: The best thing is to make `sign` a named argument, tracking it here: https://www.pivotaltracker.com/story/show/160848291

UlfNorell: This works at the moment because we haven't implemented [PT-160256544](https://www.pivotaltracker.com/story/show/160256544). It seems odd to document how to exploit the fact that the implementation isn't complete yet.

PR #215, opened 2018-09-30
Describe usage of on-chain objects in off-chain contracts

velzevur:
Changes related to aeternity/epoch#1592

### DIFF
--- a/channels/OFF-CHAIN.md
+++ b/channels/OFF-CHAIN.md
@@ -1010,19 +1010,19 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
 information needed for a contract call, including the contract address. The other participant co-signs
 the changes and the contract call is considered to be executed. Its results can be extracted from the calls tree in the state tree.
 
-### Contracts referring on-chain data
+### Contracts referring to on-chain data
 
 Contracts being used in channels off-chain calls have the exact same semantics
 as one being used on-chain. Because of the different environment though, those
 might have different results as the on-chain ones.
 
-Since there is no signle source of truth, each participant considers one's chain
-to be the correct one. This is essential for the trustless environment. Every
-contract call is pinned to the top of the chain, as it is seen by each
-participant. This could cause some differences in local contracts executions.
-If those can not be resolved, participants can always rely on the blockchain
-as an arbiter by using forcing of progress. Then the top is being used as it
-is seen by the Bitcoin-NG leader.
+Since there is no single source of truth, each participant considers their
+curren view of the chain to be the correct one. This is essential for the
+trustless environment. Every off-chain contract call is based on the top of
+the chain, as it is seen by each participant. This could cause some differences
+in local contracts executions. If those can not be resolved, participants can
+always rely on the blockchain as an arbiter by using forcing of progress. Then
+the top is being used as it is seen by the Bitcoin-NG leader.
 
 It is worth mentioning that both local contracts' executions and the forced
 progress ones must be fully deterministic and this implies some restrictions
@@ -1059,7 +1059,7 @@ deleted from the blockchain only after no contract uses the one to be deleted.
 This can not be enforced for off-chain contracts because there is no knowledge
 for them on-chain. Also if we were to use an on-chain contract referencing it
 by a registered name, the name on-chain could be changed to point to another
-different contract. This opens a security whole especially if one of the
+different contract. This opens a security hole especially if one of the
 participants is in control of the name.
 For these reasons off-chain contracts are not allowed to use on-chain ones,
 not even stateless on-chain contracts. Participants can still use well-known
--- a/channels/OFF-CHAIN.md
+++ b/channels/OFF-CHAIN.md
@@ -1009,3 +1009,61 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
 information needed for a contract call, including the contract address. The other participant co-signs
 the changes and the contract call is considered to be executed. Its results can be extracted from the calls tree in the state tree.
 
+### Contracts referring on-chain data
+
+Contracts being used in channels off-chain calls have the exact same semantics
+as one being used on-chain. Because of the different environment though, those
+might have different results as the on-chain ones.
+
+Since there is no signle source of truth, each participant considers one's chain
+to be the correct one. This is essential for the trustless environment. Every
+contract call is pinned to the top of the chain, as it is seen by each
+participant. This could cause some differences in local contracts executions.
+If those can not be resolved, participants can always rely on the blockchain
+as an arbiter by using forcing of progress. Then the top is being used as it
+is seen by the Bitcoin-NG leader.
+
+It is worth mentioning that both local contracts' executions and the forced
+progress ones must be fully deterministic and this implies some restrictions
+on using on-chain objects in the off-chain contracts. This is especially true
+for the `Chain` API:
+
+* `Chain.coinbase` - in off-chain calls we do not have miners and thus we can not
+  use the beneficiary of the generation. An address of `0` is returned instead.
+* `Chain.timestamp` - the time as in the latest block (either key or micro
+  block) as seen by the channel participant executing the contract call
+* `Chain.block_height` - the height of the chain as seen by the channel
+  participant executing the contract call
+* `Chain.difficulty` - the difficulty of the chain as seen by the channel
+  participant executing the contract call
+
+In off-chain calls we allow only look-ups for various on-chain objects.
+Creating new on-chain objects does not make any sense as those off-chain
+calls will not make it on-chain. The following look-ups are allowed
+* `AENS.resolve(name : string, pointer : string) : option(string)` - a name's
+  pointer data
+* `Oracle.get_answer(oracle_id : oracle(query_t, answer_t), query_id :
+  oracle_query(query_t, answer_t)) : option(answer_t)` - and oracle's answer
+  to a specific query
+* `Oracle.get_question(oracle_id : oracle(query_t, answer_t), query_id :
+  oracle_query(query_t, answer_t)) : query_t` - and oracle's query
+  question
+
+Registration and updates of names or asking of oracles is impossible in the
+off-chain contract calls.
+
+Using on-chain contracts in off-chain ones is a tricky task. On-chain
+contracts reference count contracts that are using to them. They can be
+deleted from the blockchain only after no contract uses the one to be deleted.
+This can not be enforced for off-chain contracts because there is no knowledge
+for them on-chain. Also if we were to use an on-chain contract referencing it
+by a registered name, the name on-chain could be changed to point to another
+different contract. This opens a security whole especially if one of the
+participants is in control of the name.
+For these reasons off-chain contracts are not allowed to use on-chain ones,
+not even stateless on-chain contracts. Participants can still use well-known
+contracts that are present on-chain but they have to copy them in their
+off-chain state. That way participants take care for their own data in a
+trustless manner - they don't have to rely on other entities keeping contracts
+on-chain for them.
+
--- a/channels/OFF-CHAIN.md
+++ b/channels/OFF-CHAIN.md
@@ -36,6 +36,7 @@
 - [Contracts](#contracts)
   + [On-chain enforcement](#on-chain-enforcement)
   + [Lifecycle](#contracts-lifecycle)
+  + [Referring on-chain objects](#contracts-referring-on-chain-data)
 
 
 ## Overview
--- a/channels/OFF-CHAIN.md
+++ b/channels/OFF-CHAIN.md
@@ -970,7 +970,7 @@ channel operations.
 Execution of a contract inside a state channel requires parties to be able to
 initialise a virtual machine to run their smart contracts in.
 
-Contracts are executed in rounds, which are denoted by `round`-s.
+Contracts are executed in rounds, which are denoted by `round` attribute.
 
 Every party executes each smart contract locally and checks if the signed states
 they receive match up with theirs. In the case that states and signatures are
--- a/channels/OFF-CHAIN.md
+++ b/channels/OFF-CHAIN.md
@@ -37,6 +37,7 @@
   + [On-chain enforcement](#on-chain-enforcement)
   + [Lifecycle](#contracts-lifecycle)
   + [Referring on-chain objects](#contracts-referring-on-chain-data)
+  + [Gas consumption](#gas-consumption)
 
 
 ## Overview
@@ -969,13 +970,14 @@ channel operations.
 Execution of a contract inside a state channel requires parties to be able to
 initialise a virtual machine to run their smart contracts in.
 
-Contracts are executed in rounds, which are denoted by rounds.
+Contracts are executed in rounds, which are denoted by `round`-s.
 
 Every party executes each smart contract locally and checks if the signed states
 they receive match up with theirs. In the case that states and signatures are
 valid, they apply the update and then send out their signature. If there was an
 error in either the execution of the contract or the signature does not match
-the state, they send a new update with the prior state but an increased round number—to avoid confusion–and their signature over it, to signal that something
+the state, they send a new update with the prior state but an increased round 
+number—to avoid confusion–and their signature over it, to signal that something
 went wrong.
 
 When operating in co-signing mode, contracts might need to be written in a way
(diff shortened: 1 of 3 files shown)


channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,61 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
 
+### Contracts referring on-chain data
+
+Contracts being used in channels off-chain calls have the exact same semantics
+as one being used on-chain. Because of the different environment though, those
+might have different results as the on-chain ones.
+
+Since there is no signle source of truth, each participant considers one's chain
+to be the correct one. This is essential for the trustless environment. Every
+contract call is pinned to the top of the chain, as it is seen by each
```
velzevur: @knarz currently each participants treats its own chain as the only source of truth. No block hash is propageted between parties. This is to change as described by PT [SC calls: pin enviroment to a block hash](https://www.pivotaltracker.com/story/show/160843011) - each off-chain call will include the on-chain block hash it was based upon.
This will impact only off-chain FSM communication between parties and will have no impact on on-chain forcing of progress. In that case the miner's top block is considerered the pinned one.

Since this will not impact consus, it is marked as post main net.

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,61 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+  oracle_query(query_t, answer_t)) : option(answer_t)` - and oracle's answer
+  to a specific query
+* `Oracle.get_question(oracle_id : oracle(query_t, answer_t), query_id :
+  oracle_query(query_t, answer_t)) : query_t` - and oracle's query
+  question
+
+Registration and updates of names or asking of oracles is impossible in the
+off-chain contract calls.
+
+Using on-chain contracts in off-chain ones is a tricky task. On-chain
```
velzevur: @knarz please confirm this is how we want it. The following paragraphs describe how it is currently implemented.

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,61 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+contracts reference count contracts that are using to them. They can be
+deleted from the blockchain only after no contract uses the one to be deleted.
+This can not be enforced for off-chain contracts because there is no knowledge
+for them on-chain. Also if we were to use an on-chain contract referencing it
+by a registered name, the name on-chain could be changed to point to another
```
knarz: How's this any different from using a name in any other context? The pointers can change, no matter what.

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,61 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+* `Oracle.get_question(oracle_id : oracle(query_t, answer_t), query_id :
+  oracle_query(query_t, answer_t)) : query_t` - and oracle's query
+  question
+
+Registration and updates of names or asking of oracles is impossible in the
+off-chain contract calls.
+
+Using on-chain contracts in off-chain ones is a tricky task. On-chain
+contracts reference count contracts that are using to them. They can be
+deleted from the blockchain only after no contract uses the one to be deleted.
```
knarz: The same is true for oracles and names as well, is it not?

channels/OFF-CHAIN.md
knarz: Pulling state into the channel versus referencing it, comes with the same caveats that you have in C, I suppose. Copying is both slow and costly while just referencing it exposes you to the risk of someone messing with your memory—in the future anyway, since the reference was valid at a given blockheight, which could be included, to make it future proof but would also increase the burden of validation.

With that in mind, I suppose it's better to just copy whatever you might need for execution.

channels/OFF-CHAIN.md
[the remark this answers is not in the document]
velzevur: _rephrased_

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,61 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+* `Chain.coinbase` - in off-chain calls we do not have miners and thus we can not
+  use the beneficiary of the generation. An address of `0` is returned instead.
+* `Chain.timestamp` - the time as in the latest block (either key or micro
+  block) as seen by the channel participant executing the contract call
+* `Chain.block_height` - the height of the chain as seen by the channel
+  participant executing the contract call
+* `Chain.difficulty` - the difficulty of the chain as seen by the channel
+  participant executing the contract call
+
+In off-chain calls we allow only look-ups for various on-chain objects.
```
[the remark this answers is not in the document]
velzevur: Not necessarily, it is up to participants when they want to make a call. This call could as well be refused by the other party.
This is to be greatly improved in [SC calls: pin enviroment to a block hash](https://www.pivotaltracker.com/story/show/160843011) as block hash of the block the call had been based on will be sent between parties (yet, not part of the on-chain protocol)

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,61 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+  pointer data
+* `Oracle.get_answer(oracle_id : oracle(query_t, answer_t), query_id :
+  oracle_query(query_t, answer_t)) : option(answer_t)` - and oracle's answer
+  to a specific query
+* `Oracle.get_question(oracle_id : oracle(query_t, answer_t), query_id :
+  oracle_query(query_t, answer_t)) : query_t` - and oracle's query
+  question
+
+Registration and updates of names or asking of oracles is impossible in the
+off-chain contract calls.
```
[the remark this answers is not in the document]
velzevur: I have to double check this, but my assumption is yes.

channels/OFF-CHAIN.md
velzevur: Yes, indeed

channels/OFF-CHAIN.md
velzevur: Both oracles and names have a various types of different timers for the participants to react upon before they disappear. That's why we have the ref counting for contracts - if another one relies on the to be deleted one - we postpone the deletion. This mechanics does not work with off-chain contracts so we lose this security there and I think it is worth mentioning it.

channels/OFF-CHAIN.md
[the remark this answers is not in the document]
knarz: Just wanted to know if it is currently used, since I'd prefer it to be at least recommended in the off-chain protocol

velzevur:
Updated `beneficiary` description according to implementation

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,61 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+  block) as seen by the channel participant executing the contract call
+* `Chain.block_height` - the height of the chain as seen by the channel
+  participant executing the contract call
+* `Chain.difficulty` - the difficulty of the chain as seen by the channel
+  participant executing the contract call
+
+In off-chain calls we allow only look-ups for various on-chain objects.
+Creating new on-chain objects does not make any sense as those off-chain
+calls will not make it on-chain. The following look-ups are allowed
+* `AENS.resolve(name : string, pointer : string) : option(string)` - a name's
```
lucafavatella: Return is [not necessarily `string`](https://github.com/aeternity/protocol/blame/25fde43eb9e958f677850e4c54228f71ebf16e16/contracts/sophia.md#L457).

May I suggest reducing the level of detail of this section, by not mentioning the whole Sophia builtin signature but rather the name, like:
> * `AENS.resolve`
> ...

(BTW I expect what matters is actually the primops for a certain VM version - rather than the Sophia language builtins - so these are indicative anyway.)

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,56 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+
+It is worth mentioning that both local contracts' executions and the forced
+progress ones must be fully deterministic and this implies some restrictions
+on using on-chain objects in the off-chain contracts. This is especially true
+for the `Chain` API:
+
+* `Chain.coinbase` - the beneficiary of the latest block (either key or micro
+  block) as seen by the channel participant executing the contract call
+* `Chain.timestamp` - the time as in the latest block (either key or micro
+  block) as seen by the channel participant executing the contract call
```
lucafavatella: I see current doc of this kind of operations is defined:
* Coarsely in [Sophia primitives](https://github.com/aeternity/protocol/blob/25fde43eb9e958f677850e4c54228f71ebf16e16/contracts/sophia.md#contract-primitives)
* In a detailed manner in [VM doc](https://github.com/aeternity/protocol/blob/25fde43eb9e958f677850e4c54228f71ebf16e16/contracts/aevm.md#chain-specific-instructions)

It looks like this list of items (Chain.coinbase, ..., Chain.difficulty) may become another documentation point to keep consistency of. May I suggest to loosen the definition in this doc and detail the VM doc. Sample for this doc:
> This is especially true for the chain-related primitives (e.g. coinbase, timestamp, block height, difficulty). Please refer to the documentation of the applicable VM version.

lucafavatella: Please detail https://github.com/aeternity/protocol/blame/9f305aea0bde1bfc34b9f23aa056206d0b97dbde/contracts/aevm.md#L55

lucafavatella: Please confirm all section https://github.com/aeternity/protocol/blob/25fde43eb9e958f677850e4c54228f71ebf16e16/contracts/aevm.md#chain-specific-instructions up to date.

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,56 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+calls will not make it on-chain. The following look-ups are allowed
+* `AENS.resolve` - a name's pointer data
+* `Oracle.get_answer` - and oracle's answer to a specific query
+* `Oracle.get_question` - and oracle's query question
+
+Registration and updates of names or asking of oracles is impossible in the
+off-chain contract calls.
+
+Using on-chain contracts in off-chain ones is a tricky task. On-chain
+contracts reference count contracts that are using to them. They can be
```
uwiger: reference-count contracts that refer to them.

channels/OFF-CHAIN.md
uwiger: only once they are no longer referenced by any other contract.

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,56 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+deleted from the blockchain only after no contract uses the one to be deleted.
+This can not be enforced for off-chain contracts because there is no knowledge
+for them on-chain. Also if we were to use an on-chain contract referencing it
```
uwiger: knowledge _of_...

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,56 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+by a registered name, the name on-chain could be changed to point to another
+different contract. This opens a security hole especially if one of the
+participants is in control of the name.
+For these reasons off-chain contracts are not allowed to use on-chain ones,
+not even stateless on-chain contracts. Participants can still use well-known
+contracts that are present on-chain but they have to copy them in their
```
uwiger: copy them _into_...

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,56 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
...
+off-chain state. That way participants take care for their own data in a
```
uwiger: take care _of_...

channels/OFF-CHAIN.md
```diff
@@ -1009,3 +1010,56 @@ contract update](../serializations.md#channel-off-chain-update-call-contract). I
 information needed for a contract call, including the contract address. The other participant co-signs
 the changes and the contract call is considered to be executed. Its results can be extracted from the calls tree in the state tree.
 
+### Contracts referring to on-chain data
+
+Contracts being used in channels off-chain calls have the exact same semantics
```
lucafavatella: I see [the “off-chain updates” are](https://github.com/aeternity/protocol/blob/1c5b17b44da6ecd661bac0fe86c2b1c935f99e8d/serializations.md#channel-off-chain-update): transfer, deposit, withdrawal, create contract, call contract.

I see [all off-chain updates’ **serializations** are part of consensus](https://github.com/aeternity/protocol/blame/1c5b17b44da6ecd661bac0fe86c2b1c935f99e8d/serializations.md#L692-L701).

I see that [the **semantics** of only the off-chain update call contract is part of consensus](https://github.com/aeternity/protocol/blame/1c5b17b44da6ecd661bac0fe86c2b1c935f99e8d/channels/ON-CHAIN.md#L468-L471). As such, its semantics needs to be detailed. I see this semantics is described in two places:

* This paragraph.

  > Contracts being used in channels off-chain calls have the exact same semantics as one being used on-chain. Because of the different environment though, those might have different results as the on-chain ones.

* [Channels ON-CHAIN document](https://github.com/aeternity/protocol/blame/1c5b17b44da6ecd661bac0fe86c2b1c935f99e8d/channels/ON-CHAIN.md#L468-L471).

  > The update MUST be a call to a contract. The caller of this update MUST be the
  > poster of the force progress transaction. `gas` and `gas_price` are
  > specified in the update as well. The gas fees are going to be paid by the poster
  > of the transaction.

Recommendations:

* I suggest to amend this paragraph with the link to the referred semantics, i.e.

  > Contracts being used in channels off-chain calls have the exact same semantics as [one being used on-chain](/contracts/contract_transactions.md#contract-call-transaction). Because of the different environment though, those might have different results as the on-chain ones.
* The [section in the ON-CHAIN document](https://github.com/aeternity/protocol/blame/1c5b17b44da6ecd661bac0fe86c2b1c935f99e8d/channels/ON-CHAIN.md#L468-L471) shall be integrated mentioning that [amount](https://github.com/aeternity/protocol/blame/1c5b17b44da6ecd661bac0fe86c2b1c935f99e8d/serializations.md#L681) is specified in the update as well, and what the semantics of this movement of coins is (there are off-chain on-chain trees, plus only the accounts of the initiator and responder “make it” as side effect of the force progress - plus state_hash/round of course).

lucafavatella: Re semantics of the off-chain upadte call contract (part of consensus) it seems it would be beneficial to document:
* Semantics re [gas](https://github.com/aeternity/epoch/pull/1592#discussion_r222615104)
* Semantics re [amount](https://github.com/aeternity/epoch/pull/1592#discussion_r222622621)

lucafavatella: And it looks like the charging of the amount depends also on the channel reserve in the channel object in the state tree.

channels/OFF-CHAIN.md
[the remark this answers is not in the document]
lucafavatella: I understand - based on [code inspection](https://github.com/aeternity/epoch/pull/1592#discussion_r222385936) - that contract execution will proceed though only using off-chain trees.

velzevur:
I believe I've addressed all comments, most notably:
* Described [what happens](https://github.com/aeternity/protocol/pull/215/files#diff-4fbc1ee2b0f1c1c88eeeafc062a7989cR505) when an exception occurs while forcing of progress
* Described [force progress gas consumption](https://github.com/aeternity/protocol/pull/215/files#diff-0eae56fb99bf9a41deeb4fd98e3e77faR1060)
* Updated AEVM's description regarding [off-chain built-ins](https://github.com/aeternity/protocol/pull/215/files#diff-1939a8160703e47dd9ef11d93cbcb168R50)

channels/OFF-CHAIN.md
```diff
@@ -968,13 +970,14 @@ channel operations.
 Execution of a contract inside a state channel requires parties to be able to
 initialise a virtual machine to run their smart contracts in.
 
-Contracts are executed in rounds, which are denoted by rounds.
+Contracts are executed in rounds, which are denoted by `round`-s.
```
uwiger: by the `round` attribute

PR #216, opened 2018-10-01
update burner contract for token migration

nikita-fuchs:


GENESIS.md
```diff
@@ -101,19 +100,61 @@ contract tokenFallback {
...
+                          //2. store only those which havn't been stored before, which would require a seperate mapping + array to iterate over, so I suggest we go with option 1.
+    }
+    
+    
+    // Store the hashes of Aadmins' msg.data used for counting up the AE delivery batch
+    mapping (address => bytes32) private multiSigHashes;
+    // Keep track of token burn batches - use this number for filtering in the emitted Burn event.
+    uint16 public AEdeliveryBatchCounter = 0;
+    
+    // The admins which may count up the AE delivery batch count
```
knarz: Do we need a multisig here? A rogue admin could only ever increment the batch number and not wreak any havoc that way—unless we consider the possibility of overflowing the counter.

I think I'm a bit more concerned with one private key going missing and then not being able to increment batch number.

What do you think?

emil-apeunit: I think so too, let's skip multisig because no real harm can be done, but having access when we need it is essential.

nikita-fuchs:
agree.

evdoks:
@nikita-fuchs, @knarz,  is there something else that should be done here?

nikita-fuchs:
@knarz is that the latest version above (the one you reviewed) ? Can't
really tell on the phone.

Sergei Evdokimov <notifications@github.com> schrieb am Mo., 29. Okt. 2018,
09:52:

PR #219, opened 2018-10-04
PT-160981321 Add intended usage of mining API

jur0:
[PT-160981321](https://www.pivotaltracker.com/n/projects/2124891/stories/160981321)

### DIFF
--- a/epoch/api/mining_api_usage.md
+++ b/epoch/api/mining_api_usage.md
@@ -1,0 +1,68 @@
+[back](./README.md)
+# Mining - intended usage
+
+## Introduction
+
+In order to get a reward and transaction fees, the epoch node needs to mine
+a key block. With increased difficulty, the probability of mining a key block
+is lower. Therefore, the miners can form mining pools where they share the
+processing power over the network so they can mine key blocks more often. They
+split the reward according to the amount of work they contribute to mine a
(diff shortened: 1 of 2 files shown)


epoch/api/mining_api_usage.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Epoch node modes
+
+### Pool manager
+
+The epoch node can be configured (using `autostart: false`) not to mine key
+blocks. Even though the node does not mine the key blocks, it still produces
+the key block candidates that do not have `nonce` and `pow` set. These key
+block candidates can be retrieved using the
+[/key-blocks/pending endpoint](https://aeternity.github.io/epoch-api-docs/?config=https://raw.githubusercontent.com/aeternity/epoch/master/apps/aehttp/priv/swagger.json#/external/GetPendingKeyBlock):
```
dincho: Not sure it's OK pointing to `master` docs here. It might get out of sync.

jur0: I think all the HTTP API related links point to the master. Alternatively we can point to the latest release, the link would be: https://aeternity.github.io/epoch-api-docs/#/external/GetPendingKeyBlock or we can remove the HTTP API links from the docs. It'd be done in a separate PR though.

PR #223, opened 2018-10-08
Abort is now a built-in function and should not be defined by user

rvirding:
This is refers to [PT-160993470](https://www.pivotaltracker.com/story/show/160993470)

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -513,7 +513,7 @@ The block-chain environment available to a contract is defined in three name spa
 
 ### Exceptions
 
-Contracts can fail with an (uncatchable) exception using the function
+Contracts can fail with an (uncatchable) exception using the built-in function
 
 ```
 abort(reason : string) : 'a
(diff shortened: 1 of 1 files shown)


contracts/sophia.md
```diff
@@ -513,7 +513,7 @@ The block-chain environment available to a contract is defined in three name spa
 
 ### Exceptions
 
-Contracts can fail with an (uncatchable) exception using the function
+Contracts can fail with an (uncatchable) exception using the built-in function
 
 ```
 abort(reason : string) : 'a
```
lucafavatella: What does `'a` mean? Maybe it shall be [tuple of zero elements](https://github.com/aeternity/epoch/pull/1623/files#diff-188efb931c56a6fb0aa1263174fc8eabR108)?

rvirding: That I don't know, it was already there. It is a type variable so it could be what I want it to be.  As far as I know there is no `no_return` type which would be the best.  Having it as `()` would be better and match the current implementation. I will fix it.

rvirding: Fixed it so docs say it return `()`.

lucafavatella:
It goes with https://github.com/aeternity/epoch/pull/1623 - merged.

PR #226, opened 2018-10-10
Document transaction gas

jur0:
[PT-161115029](https://www.pivotaltracker.com/n/projects/2124891/stories/161115029)


lucafavatella:
I understand this goes alongside https://github.com/aeternity/epoch/pull/1630 - ok.

PR #229, opened 2018-10-11
Clarify channel ports

velzevur:


epoch/api/README.md
```diff
@@ -31,6 +31,13 @@ The epoch node exposes the following APIs:
     * It is defined in the rest of this document;
     * It is meant to be exposed on the Internet;
     * Its TCP port is configurable.
+  * Range of external Channels noise endpoints. It consists of as many TCP
+    endpoints as needed.
+    * They are encrypted and authenticated using the Noise protocol;
```
lucafavatella: Where is it defined? Via Swagger schema? In the rest of this document?

epoch/api/README.md
```diff
@@ -31,6 +31,13 @@ The epoch node exposes the following APIs:
     * It is defined in the rest of this document;
     * It is meant to be exposed on the Internet;
     * Its TCP port is configurable.
+  * Range of external Channels noise endpoints. It consists of as many TCP
```
lucafavatella: Which Noise public key is used by the listener?
Which Noise public key of the other party does the listener expect/validate?

velzevur: [Addressed](https://github.com/aeternity/protocol/pull/229/commits/a086114ae96dd9f500cbc6e03869d5ee3a2b0502#diff-8b16649bc0188ba5b35d5f92f6b8656aR40)

epoch/api/README.md
```diff
@@ -31,6 +31,15 @@ The epoch node exposes the following APIs:
...
```
lucafavatella: Could you also please clarify https://github.com/aeternity/epoch/blob/master/docs/configuration.md#channels for next release?

epoch/api/README.md
```diff
@@ -31,6 +31,15 @@ The epoch node exposes the following APIs:
     * It is defined in the rest of this document;
     * It is meant to be exposed on the Internet;
     * Its TCP port is configurable.
```
uwiger: Also, the listen IP address is configurable (default: soon 127.0.0.1, but currently 0.0.0.0, meaning that the node listens on all available IP interfaces.)

epoch/api/README.md
lucafavatella: Where is the schema of its payload defined?

velzevur: Clarified [in epoch](https://github.com/aeternity/epoch/pull/1667)

lucafavatella: I see addressed (`/channels/OFF-CHAIN.md#messages`). Thanks.

velzevur:
The fragmentation is already tracked as [state channels should be able to reuse listen sockets](https://www.pivotaltracker.com/story/show/161143845). I've created a PT [SC WebSocket Versioning](https://www.pivotaltracker.com/story/show/161288395)

velzevur:
@dincho - this is to be changed by [state channels should be able to reuse listen sockets](https://www.pivotaltracker.com/story/show/161143845)

PR #235, opened 2018-10-15
Describe result of off-chain calls

velzevur:


contracts/aevm.md
```diff
@@ -43,6 +43,10 @@ The following arguments are encoded as Sophia data.
 
 Note that the gas cost indicated in the table above does not include the gas required for the call instruction to the primop.
 
+Calling state changing prim-ops in oracles, names and channel's trees in
```
lucafavatella: Are there any primops re channel's tree?

velzevur: not yet. If those happen (and this is before we have virtual state channels), those must be throwing that error as well

PR #240, opened 2018-10-16
Pt 161243489 json rpc msg logs

uwiger:
See [PT #161243489](https://www.pivotaltracker.com/story/show/161243489)


lucafavatella:
I see related code merged as https://github.com/aeternity/epoch/pull/1662

PR #241, opened 2018-10-17
removes lock from channel in open state

knarz:
This PR removes the lock period from channels while they are open and instead keeps track of two different round numbers. It also tries to clarify handling of disputes in the different states of a channel.

### DIFF
--- a/channels/ON-CHAIN.md
+++ b/channels/ON-CHAIN.md
@@ -1,46 +1,51 @@
 # On-chain
 
-- [Establishment](#establishing-channel-on-chain)
-	- [`channel_create`](#channel_create)
-- [Updates](#updating-channel-on-chain)
-	- [`channel_deposit`](#channel_deposit)
-	- [`channel_withdraw`](#channel_withdraw)
-	- [`channel_snapshot_solo`](#channel_snapshot_solo)
-- [Closing](#closing-channel-on-chain)
-	- [`channel_close_mutual`](#channel_close_mutual)
-	- [`channel_close_solo`](#channel_close_solo)
-	- [`channel_slash`](#channel_slash)
-	- [`channel_settle`](#channel_settlement)
-- [Force progress](#force-progress-on-chain)
+Operating a state channel requires, at least, an initial setup via a
+`channel_create` transaction, that locks up a configurable amount of coins from
+each involved party in the channel. This requires consent, in the form of
+signatures, from all participants.
 
-Each party keeps a state tree specific for the channel. It consists of all the
-channel data: accounts, contracts and etc. and has the same structure as the
-on-chain state tree. Off-chain transactions update this channel auxiliary tree.
-It is a responsibility of the parties to keep this locally. Solo closing
-transactions provide a proof of inclusion for it instead of
-posting the whole tree.
-Each off-chain update consists of updates being applied on top of channel state
-tree and an integer value `round` representing when it happened. Since `round`
-must always be bumped, provided two off-chain transactions we can reason which
-was performed earlier than the other.
+In the ideal case, all on-chain transactions get signed off by all participants,
+implying that there are no disagreements. These operations can be committed
+immediately and will generally override all operations initiated unilaterally.
 
-Each on-chain updating transaction provides two fields that are essential for
-future conflict resolution: `round` and `state_hash`. The state hash is the
-root hash of the channel state tree after the on-chain has been applied to the
+Transactions not signed off by everyone, except the `channel_snapshot_solo`, can
+be disputed via `channel_slash` and `channel_force_progress_tx` transactions.
+During normal operation, these solo transactions can be disputed indefinitely.
+If the channel is in the closing state, disputes have to happen within the
+pre-negotiated `lock_period`, given that closure requires a bounded finality.
+
+Each transaction updating on-chain state provides two fields essential for
+future conflict resolution: `round` and `state_hash`. The state hash is the root
+hash of the channel's state tree after the on-chain has been applied to the
 local state tree. The `round` is the _next_ state channel internal round. Thus
 the on-chain update transaction represents on-chain the next off-chain state
 of the channel. This way we can solo close a channel according to the last
 on-chain state. All we have to do is to provide a proof of inclusion having
 the same `state_hash`.
 
-## Establishing a channel on-chain
+
+- [Establishing a channel](#establishing-a-channel)
+	+ [`channel_create`](#channel_create)
+- [Updating a channel](#updating-a-channel)
+	+ [`channel_deposit`](#channel_deposit)
+	+ [`channel_withdraw`](#channel_withdraw)
+	+ [`channel_snapshot_solo`](#channel_snapshot_solo)
+- [Closing a channel](#closing-a-channel)
+	+ [`channel_close_mutual`](#channel_close_mutual)
+	+ [`channel_close_solo`](#channel_close_solo)
+	+ [`channel_settle`](#channel_settlement)
+- [Disputing updates](#disputing-updates)
+	+ [`channel_slash`](#channel_slash)
+- [Forcing progress](#forcing-progress)
+	+ [`channel_force_progress_tx`](#channel_force_progress_tx)
+- [Channel state tree](#channel-state-tree)
+
+## Establishing a channel
 
 All of the on-chain operations could be submitted by any peer but we assume that
 the initiating peer pays for the setup of the channel. Therefore the `initiator`
 MUST pay the standard transaction fees.
-
-(***TODO***: should fees be directly be deducted from channel balance?)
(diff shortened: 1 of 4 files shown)


channels/ON-CHAIN.md
```diff
@@ -542,3 +546,172 @@ Since the miner is expending resources for the contract's execution, the gas
...
+Each block MUST commit to a Merkle Patricia tree of open channels, where the
+`channel_id` specifies the path.
+At a leaf, nodes store information pertaining to the current state of the given
+channel.
+
+- `channel_id`
+- `initiator_pubkey`
+- `responder_pubkey`
+- `total_amount`
+- `initiator_amount`
```
lucafavatella: I see that [serialization doc](https://github.com/aeternity/protocol/blob/2d3246e9bed457f4399e4dde1b3e29ee3af11914/serializations.md#channel) (linked below) has `responder_amount` while absent here. Any reasons?

channels/ON-CHAIN.md
```diff
@@ -542,3 +546,172 @@ Since the miner is expending resources for the contract's execution, the gas
...
+- `channel_reserve`
```
lucafavatella: I would expect the state shall include the delegates - for the slash transaction. That field is present in the [serialization doc](https://github.com/aeternity/protocol/blob/2d3246e9bed457f4399e4dde1b3e29ee3af11914/serializations.md#channel).

channels/ON-CHAIN.md
```diff
@@ -542,3 +546,172 @@ Since the miner is expending resources for the contract's execution, the gas
...
+- `state_hash`: last published state_hash
+- `round`: last known co-signed round
+- `solo_round`: last round produced via a `channel_force_progress_tx`
```
lucafavatella: I find no traces of this in the code. Which ticket is meant to track this implementation? [This](https://www.pivotaltracker.com/story/show/160613339) cc @velzevur ?

I see this field not in the [serialization doc](https://github.com/aeternity/protocol/blob/2d3246e9bed457f4399e4dde1b3e29ee3af11914/serializations.md#channel). I expect this to be left for update contextually to the implementation ticket.

channels/ON-CHAIN.md
```diff
@@ -1,47 +1,53 @@
...
+- [Closing a channel](#closing-a-channel)
+	+ [`channel_close_mutual`](#channel_close_mutual)
+	+ [`channel_close_solo`](#channel_close_solo)
+	+ [`channel_settle`](#channel_settlement)
+- [Disputing updates](#disputing-updates)
+	+ [`channel_slash`](#channel_slash)
+- [Forcing progress on-chain](#forcing-progress)
+	+ [`channel_force_progress_tx`](#channel_force_progress_tx)
+
+- [Channel state tree](#channel_state_tree)
```
lucafavatella: s/_/-/g

channels/ON-CHAIN.md
velzevur: Yes, this `solo_round` is a new functionality to be added. This PR describes how it shall work

channels/ON-CHAIN.md
knarz: I was going to address that in a separate PR but can include it in this one as well

knarz: I adjusted the PR to include and update to the serializations

channels/ON-CHAIN.md
lucafavatella: I see added in dc3fa1b. Thanks.

channels/ON-CHAIN.md
lucafavatella: I see addressed in dc3fa1b. Thanks.

channels/ON-CHAIN.md
```diff
@@ -60,18 +65,27 @@ Channel(cid).amount := initiator_amount + responder_amount
...
-  solo closing sequence: except for the participants of the channel, they are
-  the only ones that can provide a slash transaction.
-- `state_hash`: the root hash of the channel state tree; This is not validated, just kept in the channel's object
-- `nonce`
+- `initiator_id`: account id of the initiating peer
+- `responder_id`: account id of the responding peer
+- `initiator_amount`: unsigned amount of coins the initiator commits to the
+  channel
+- `responder_amount`: unsigned amount of coins the responder commits to the
+  channel
```
lucafavatella: @knarz (Potentially in separate new PR) Could you please detail semantics of `channel_reserve` in `/channels/ON-CHAIN.md#channel_withdraw` ?

I see the requirements below mention `channel_reserve` but that is not listed/described among fields here. This may surprise reader. (It is present in the linked serializations.md ) I tried to understand semantics of `channel_reserve` from the rest of the document, and I would have expected it to be mentioned in `channel_withdraw` though I did not find it there. Hence my request.

@velzevur @uwiger
* The withdraw tx [checks the double of the reserve against one combined amount ("channel amount" - not sure of its semantics)](https://github.com/aeternity/epoch/blob/f0493f6eb0e2fec4fabe639da944e7153f4ebb89/apps/aechannel/src/aesc_withdraw_tx.erl#L291-L292), rather than checking the reserve against each single party (initiator and responder) as the create tx does. Is this intended?
* The force progress tx [does not check that each single party (initiator and responder) has at least channel reserve in its amount](https://github.com/aeternity/epoch/blob/f0493f6eb0e2fec4fabe639da944e7153f4ebb89/apps/aechannel/src/aesc_utils.erl#L522). Is this intended?
  * If introducing such check we may consider also to rather remove such unspendable reserve from each account (if it is an unspendable amount per party) *before* running the contract and re-adding it afterwards - so to prevent contract to cause miner to spend computational effort with those coins while it should not have done so.

channels/ON-CHAIN.md
```diff
@@ -1,47 +1,52 @@
...
 the on-chain update transaction represents on-chain the next off-chain state
 of the channel. This way we can solo close a channel according to the last
 on-chain state. All we have to do is to provide a proof of inclusion having
 the same `state_hash`.
 
-## Establishing a channel on-chain
+
+- [Establishing a channel](#establishing-a-channel)
+	+ [`channel_create`](#channel_create)
+- [Updating a channel](#updating-a-channel)
```
lucafavatella: `#updating-a-channel-on-chain`
Or you meant to change section header?

channels/ON-CHAIN.md
velzevur: #### Withdrawal tx
Since in the withdraw transaction it is only the `state_hash` of the new off-chain state tree that is being provided, we can not enforce that both participants have at least `channel_reserve` tokens left in their off-chain balances. Thus we can not enforce this further than the on-chain total channel amount never falls lower than the double `channel_reserve` amount.
Alternative would be providing a proof of inclusion for both account's balances (or a full set of trees) in the withdrawal transaction. This way we could validate amounts.
I'd rather go with the current _optimistic check_ approach

#### Force progress
Force progress includes a shift of tokens from the caller to the off-chain contract. A force progress transaction then only potentially lowers the amount of tokens possesed by the `caller` (forcer of progress) and should not validate the other participant. The `aesc_utils:process_force_progress/7` indeed does check the [caller of the contract doesn't spend too much](https://github.com/aeternity/epoch/blob/f0493f6eb0e2fec4fabe639da944e7153f4ebb89/apps/aechannel/src/aesc_offchain_update.erl#L268-L273) but this is burried deep in the channel internals.

lucafavatella: > indeed does check the caller of the contract doesn't spend too much

I am not sure about the following case though did not arrive at checking. If:
* Parties start at A=20 and B=40, and reserve is 10 (hence 2*10).
* There is withdrawal of 40 from B -> valid (I am ignoring fees) because leaving 20 that is > 2*10.
* Then there is force progress that removes 10 from A -> valid because A (was 20 and) becomes 10. BUT the total amount is less than 2*10 (10 + 0).

velzevur: yes, this is possible, indeed

knarz: We're going to address the `channel_reserve` in a separate PR

channels/ON-CHAIN.md
```diff
@@ -542,3 +545,174 @@ Since the miner is expending resources for the contract's execution, the gas
...
+- `total_amount`
+- `initiator_amount`
+- `responder_amount`
+- `channel_reserve`
+- `state_hash`: last published state_hash
+- `round`: last known co-signed round
+- `solo_round`: last round produced via a `channel_force_progress_tx`
+- `lock_period`: agreed upon locking period by peers
+- `closes_at`: on-chain channel closing height
+- `locked_until`: on-chain channel height after which the channel can be settled
```
velzevur: What is the difference between 
```
`locked_until`: on-chain channel height after which the channel can be settled
```
and 
```
 `closes_at`: on-chain channel closing height
```

PR #247, opened 2018-10-22
PT-161380219: Add check to disallow reentrant calls for Sophia EVM

UlfNorell:
See https://github.com/aeternity/epoch/pull/1679

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -93,6 +93,9 @@ without calling it you can write
     Chain.spend(v.address, amount)
 ```
 
+Note that reentrant calls are not permitted. In other words, when calling
+another contract it cannot call you back.
+
 ### Mutable state
 
 Sophia does not have arbitrary mutable state, but only a limited form of


contracts/sophia.md
```diff
@@ -93,6 +93,9 @@ without calling it you can write
     Chain.spend(v.address, amount)
 ```
 
+Note that reentrant calls are not permitted. In other words, when calling
```
lucafavatella: Based on [code PR](https://github.com/aeternity/epoch/pull/1679/files?utf8=%E2%9C%93&diff=split&w=1#diff-a283035db2ad319c06dab29a7f43e182R156) this seems more part of [VM semantics rather than language ABI](https://github.com/aeternity/protocol/blame/a8fc667df4b777ebbb7d5edbffe377d6ce3a085a/contracts/contract_vms.md#L22). It looks like the [VM spec](https://github.com/aeternity/protocol/blob/a8fc667df4b777ebbb7d5edbffe377d6ce3a085a/contracts/aevm.md) shall mention which instructions diverge from EVM because of this and how too.

PR #252, opened 2018-10-24
Document that the return value for the call in the create transaction…

happi:
… is cleared.
See https://github.com/aeternity/epoch/pull/1688


contracts/contract_state_tree.md
```diff
@@ -66,3 +66,7 @@ See [Call Serialization](../serializations.md#contract-call) for the serializati
 Each call includes the caller address and caller nonce that, together,
 correlate the call with the transaction in the block that did the
 call.
+
+The call object in the state tree contains the return value of a contract call transaction.
+For the contract create transaction the return value of the inital call is set to
+the empty byte array.
```
velzevur: Just a suggestion: ~the~ _an_ empty byte array

PR #258, opened 2018-10-30
Update to new calling convention

gorillainduction:


lucafavatella:
I see it refers mostly to https://github.com/aeternity/epoch/pull/1702

PR #274, opened 2018-11-22
Do not burn tokens

velzevur:
Compliment to aeternity/epoch#1809


channels/ON-CHAIN.md
```diff
@@ -391,8 +393,13 @@ considered closed and allow no further modifications.
 The transaction MUST be signed using private key corresponding to the public key `from_id`.
 
 The amounts must correspond to the ones on-chain, provided by the last
-`channel_close_solo` or `channel_slash`.
+`channel_close_solo` or `channel_slash`. The sum of those final amounts form
+the total closing amount of the channel.
 
+agreed upon amounts. The total closing amount of a channel is computed bydd
```
gorillainduction: Stray duplication of text

consensus/locking.md
```diff
@@ -0,0 +1,9 @@
+Coins locking
+==
+
+We want a predictable inflation so there MUST not be any token burning. In order
```
knarz: How is this different than burning? Protocol mandated burning is visible and therefore easy to track.

PR #275, opened 2018-11-27
Set protocol version to 1

zp-sd:


consensus/consensus.md
```diff
@@ -150,8 +150,8 @@ nonce:
 #### Key Blocks
 
 ```
-PROTOCOL_VERSION: 25
-GENESIS_VERSION: 25
+PROTOCOL_VERSION: 1
```
lucafavatella: I see https://github.com/aeternity/epoch/blob/v1.0.0/apps/aecore/include/blocks.hrl#L3

PR #276, opened 2018-12-04
describe microblock confirmations

uwiger:
See [PT #162416632](https://www.pivotaltracker.com/story/show/162416632).

### DIFF
--- a/channels/OFF-CHAIN.md
+++ b/channels/OFF-CHAIN.md
@@ -332,7 +332,7 @@ As outlined in [this Bitcoin-NG blog post](http://hackingdistributed.com/2015/11
 will still wait until the transaction is seen inside a microblock. This will mean that 
 the transaction has been gossiped, picked up from the mempool and accepted by a miner.
 
-Note that high-value transactions will still need to wait for a number of keyblocks, to
+Note that high-value transactions should wait for a number of keyblocks in order to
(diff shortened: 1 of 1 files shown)


channels/OFF-CHAIN.md
```diff
@@ -325,6 +326,16 @@ different amounts, then that might communicate that it wants these instead.)
   blockchain to stop a cheater
 - `responder_pubkey` MUST be a valid ed25519 pubkey
 
+#### Microblock confirmation
+
+As outlined in [this Bitcoin-NG blog post](http://hackingdistributed.com/2015/11/09/bitcoin-ng-followup/), _an NG microblock offers strictly stronger guarantees than a 0-confirmation in Bitcoin_. In Aeternity, when the minimum-depth confirmation is set to zero, the system
+will still wait until the transaction is seen inside a microblock. This will mean that 
+the transaction has been gossiped, picked up from the mempool and accepted by a miner.
+
+Note that high-value transactions will still need to wait for a number of keyblocks, to
```
velzevur: I'd rather have this like 

> Note that high-value transactions ~will still need to~ should wait for a number of keyblocks, to

as the waiting it is not part of the protocol but rather a suggestion. This is a matter of taste and not a blocker.

PR #277, opened 2018-12-05
removes epoch API documentation

knarz:
This removes the Epoch API documentation, which was temporarily housed in here, in an effort to make the protocol implementation agnostic.

This should only be merged once the documentation is available somewhere else.


ThomasArts:
Documentation has changed too much to benefit from this PR.

PR #278, opened 2018-12-12
Reset version and improve PoW in consensus.md

hanssv:


consensus/consensus.md
```diff
@@ -150,8 +150,8 @@ nonce:
 #### Key Blocks
 
 ```
-PROTOCOL_VERSION: 25
```
lucafavatella: I see these protocol number update in master already, not sure why here. Though harmless so ok.

PR #283, opened 2018-12-17
Pt 162663022 missing and added sophia operators

hanssv:


contracts/sophia.md
```diff
@@ -767,6 +769,7 @@ In order of highest to lowest precedence.
 | Operators | Associativity
 | --- | ---
 | `!` `bnot` | right
+| `^` | left
```
lucafavatella: Re precedence I see correspondence with https://github.com/aeternity/epoch/pull/1928/commits/826ab4bf27b07d4cfb8bc4b5c64af7877a6fbbbc#diff-d896ebc1c9e8e88a88db67bcb04d717dR179

PR #291, opened 2019-01-15
OTP release rename with backward compatible bin/epoch

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/163191726


lucafavatella:
Not needed - see https://github.com/aeternity/aeternity/pull/2057

PR #292, opened 2019-01-17
Remove `uint` from the docs.

kraikov:
Remove `uint` since it's not yet present in Sophia.


lucafavatella:
@kraikov Thanks for your contribution!

PR #298, opened 2019-01-24
Rename epoch folder to node

lucafavatella:


knarz:
Why are they not just finally deleted from this repo?

lucafavatella:
> Why are they not just finally deleted from this repo?

@knarz This is an intermediate step [agreed with](https://www.pivotaltracker.com/story/show/163321177/comments/198578043) @evdoks in order not to perform too many unrelated changes during the software renaming.

PR #307, opened 2019-02-11
Bring protocol in line with real world

ThomasAerts:
Both Roma and Minerva accept name updates that have a TTL of 50000 blocks into the future.

### DIFF
--- a/AENS.md
+++ b/AENS.md
@@ -309,7 +309,7 @@ See also [mechanisms section](#mechanisms).
 The `update` transaction MUST be signed by the owner
 of the name entry to be updated.
 
-The `expire_by` MUST NOT be more than 36000 blocks into
+The `expire_by` MUST NOT be more than 50000 blocks into
 the future.
 
 `update` transaction may be used to extend the lease of the name.


AENS.md
```diff
@@ -309,7 +309,7 @@ See also [mechanisms section](#mechanisms).
 The `update` transaction MUST be signed by the owner
 of the name entry to be updated.
 
-The `expire_by` MUST NOT be more than 36000 blocks into
+The `expire_by` MUST NOT be more than 50000 blocks into
```
lucafavatella: I see correct in [1.4.0](https://github.com/aeternity/aeternity/blob/v1.4.0/apps/aens/src/aens_update_tx.erl#L247), [master](https://github.com/aeternity/aeternity/blob/dbbd2c65f89a8bf7b9c6cf3413a7c83f55104e5f/apps/aecore/src/aec_tx_processor.erl#L619), [minerva](https://github.com/aeternity/aeternity/blob/df35135ff2ad6d5f42a496c2447a59afbb1781e6/apps/aecore/src/aec_tx_processor.erl#L623).

PR #310, opened 2019-02-12
Add compiler version to serialization

ThomasArts:


lucafavatella:
I see related to https://github.com/aeternity/aeternity/pull/2146

serializations.md
```diff
@@ -835,14 +835,23 @@ The binary is a serialized Merkle Patricia Tree.
 
 The binary is a serialized Merkle Patricia Tree.
 
-#### Sophia byte code
+#### Sophia byte code (version 1, Roma release)
```
lucafavatella: Unclear if still legal post-Roma.

ThomasArts: Yes, this tag will remain to exist, but I guess as soon as we really hard fork, we restrict it.

ThomasArts: In fact it probably is a good idea to allow both for Minerva, but our software is always adding a compiler version. Then in the next hard-fork we can fully remove the possibility to add contracts to chain without a compiler version included.

PR #313, opened 2019-02-14
Fix typo

snario:


channels/README.md
```diff
@@ -27,7 +27,7 @@ level issues, that shaped our state channel design.
 - [Channel types](#channel-types)
 - [Topology](#topology)
 - [Incentives](#incentives)
-- [Artefacts](#artefacts)
+- [Artifacts](#artifacts)
```
lucafavatella: Thanks for aiming at making the documentation better!

I [read](https://en.wiktionary.org/wiki/artefact#Noun) that artefact and artifact are alternative forms, so this change would be debatable so I lean towards not including it.

PR #317, opened 2019-02-20
Merge minerva to master

lucafavatella:
Companion of https://github.com/aeternity/aeternity/pull/2188


consensus/consensus.md
```diff
@@ -211,7 +220,9 @@ MUST have valid signature from private key belonging to `sender`.
 
 Each (on-chain) transaction has the following fields:
 * Fee. It MUST be at least the gas for the transaction multiplied by the
-  minimal gas price, which is `1` (*10^-18) aeons.
+  minimal gas price, which (after MINERVA hard fork height) is
+  `1000000` (*10^-18) aeons. (Before MINERVA hard fork height it was
```
dincho: Am I reading it wrong it it has the same amount in the brackets `(*10^-18)` ?

gorillainduction: That's the denominator, so it is the same.

PR #322, opened 2019-03-13
Adjust wording in "state" definition

nikita-fuchs:
Proposing this small change in wording to make this documentation easier understandable to newcomers.


knarz:
If this is still desired maybe open a new PR with more accurate wording.

nikita-fuchs:
Why close, @knarz ? I think @lucafavatella would approve and we could just merge ? :)

PR #325, opened 2019-03-13
added table of contents

Pegahbit:


contracts/sophia.md
```diff
@@ -1,4 +1,5 @@
 [back](./contracts.md)
+[back](./contracts.md)
```
dincho: double back ?

contracts/sophia.md
```diff
@@ -9,6 +10,73 @@ to a blockchain (the Æternity BlockChain). Thus some features of conventional
 languages, such as floating point arithmetic, are not present in Sophia, and
 some blockchain specific primitives, constructions and types have been added.
 
+## Table of Contents
```
thepiwo: Is this autogenerated or will this have to be updated with every new addition? Who's in charge for keeping track of this?

contracts/sophia.md
Pegahbit: by mistake

contracts/sophia.md
sennui: ^^ maybe the script to generate it can go to top level?

Pegahbit: Not autogenerated. A better overview was much in demand. Whoever is adding something to the documentation could expand the table of contents as well?

thepiwo: History shows stuff like this will be terribly out to date after a few iterations, and be more confusing that helpful. So I'd propose to either have a script to generate this or choose an alternative frontend than GitHub for this.

Pegahbit: As far as I recognize there wasn't any script used at the "State Channels" sections as well. So who could make a script proposal? People from the community are really struggling with a bad overview and long searches.

dincho: I tend to agree with @Pegahbit that SC already have ToC here: https://github.com/aeternity/protocol/tree/master/channels

Script or different frontend could be additional improvement PR as it seems a decision of maintaining the ToC manually is already taken with SC.

contracts/sophia.md
dincho:
```suggestion
```

contracts/sophia.md
thepiwo: I agree, just noticed this, definitely can be a separate improvement

Pegahbit: So what's the status now? It's still open and not merged. Could we solve this please?

PR #340, opened 2019-04-17
Add bytes(Len) and use base58 encoded address literals

hanssv:
~~TODO: address literal syntax/scan/parse. (cc @UlfNorell )~~

Also this PR aims to further clarify the format of literals.

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -843,6 +843,13 @@ private public rec stateful switch true type record datatype
 - `Bytes = #[0-9A-Fa-f]+` byte array literal
 - `String` string literal enclosed in `"` with escape character `\`
 - `Char` character literal enclosed in `'` with escape character `\`
+- `AccountAddress` base58-encoded 32 byte account pubkey with `ak_` prefix
+- `ContractAddress` base58-encoded 32 byte contract address with `ct_` prefix
+- `OracleAddress` base58-encoded 32 byte oracle address with `ok_` prefix
+- `OracleQueryId` base58-encoded 32 byte oracle query id with `oq_` prefix
+
+See the [identifier encoding scheme](./node/api/api_encoding.md) for the
+details on the base58 literals.
 
 ### Layout blocks
 
@@ -1000,6 +1007,8 @@ Expr ::= '(' Args ')' '=>' Block(Stmt)      // Anonymous function    (x) => x + 
        | '(' Expr ')'                       // Parens                (1 + 2) * 3
        | Id | Con | QId | QCon              // Identifiers           x, None, Map.member, AELib.Token
        | Int | Bytes | String | Char        // Literals              123, 0xff, #00abc123, "foo", '%'
+       | AccountAddress | ContractAddress   // Chain identifiers
+       | OracleAddress | OracleQueryId      // Chain identifiers
 
 FieldUpdate ::= Path '=' Expr
 Path ::= Id                 // Record field
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -17,7 +17,7 @@ some blockchain specific primitives, constructions and types have been added.
     -  [Mutable state](#mutable-state)
     -  [Namespaces](#namespaces)
     -  [Splitting code over multiple files](#splitting-code-over-multiple-files)
-    -  [Types](#types) 
+    -  [Types](#types)
     -  [Arithmetic](#arithmetic)
     -  [Bit fields](#bit-fields)
     -  [Type aliases](#type-aliases)
@@ -76,7 +76,7 @@ some blockchain specific primitives, constructions and types have been added.
     - [Return](#return)
     - [Storing the contract state](#storing-the-contract-state)
     - [Remote contract calls](#remote-contract-calls)
-    
+
 ## Language Features
 ### Contracts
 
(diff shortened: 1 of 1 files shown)


contracts/sophia.md
```diff
@@ -241,22 +241,47 @@ Sophia has the following types:
...
+| Chain.ttl  | Time-to-live (fixed height or relative to current block) | ```FixedTTL(1050)``` ```RelativeTTL(50)```
 | oracle('a, 'b)       | And oracle answering questions of type 'a with answers of type 'b |  ```Oracle.register(acct, qfee, ttl)```
 | oracle_query('a, 'b) | A specific oracle query |  ```Oracle.query(o, q, qfee, qttl, rttl)```
+| contract   | A user defined, typed, contract address | ```function call_remote(r : RemoteContract) = r.fun()```
+
+### Literals
+| Type       | Constant/Literal example(s)
+| ---------- | -------------------------------
+| int        | `-1`, `2425`, `4598275923475723498573485768`
+| address    | `ak_2gx9MEFxKvY9vMG5YnqnXWv1hCsX7rgnfvBLJS4aQurustR1rt`
```
thepiwo: I assume addresses are okay to start with other prefixes as well? 

Will there be some validation included in the VM? e.g. SpendTx only to `ak_` Contract Interfaces only using `ct_`

contracts/sophia.md
```diff
@@ -241,22 +241,47 @@ Sophia has the following types:
...
+| string     | `"This is a string"`
+| list       | `[1, 2, 3]`, `[(true, 24), (false, 19), (false, -42)]`
+| tuple      | `(42, "Foo", true)`
+| record     | `balance{ owner = Call.origin, value = 100000000 }`
+| map        | `{["foo"] = 19, ["bar"] => 42}`, `{}`
+| option('a) | `Some(42)`, `None`
+| state      | `state{ owner = Call.origin, magic_key = #a298105f }`
+| event      | `EventX(0, "Hello")`
+| hash       | `#000102030405060708090a0b0c0d0e0f000102030405060708090a0b0c0d0e0f`
+| signature  | `#000102030405060708090a0b0c0d0e0f000102030405060708090a0b0c0d0e0f000102030405060708090a0b0c0d0e0f000102030405060708090a0b0c0d0e0f`
```
thepiwo: shouldn't signature be either a `byte(X)` or a special literal to use the nodes encoding using `sg_`?

contracts/sophia.md
hanssv: The idea was to have an address being equivalent to an account, not sure you'll need a multi-purpose address?

contracts/sophia.md
hanssv: I am not sure... Since a signature is really interchangeable with `bytes(64)` it feels natural to use the same kind of literal. Also signatures might come from elsewhere (Bitcoin?) and then it does not feel natural to use Aeternity encoding :thinking:

contracts/sophia.md
thepiwo: maybe the type can then be `account`?

contracts/sophia.md
thepiwo: I do agree, I was just throwing this thought out here. But we could destinguish aeternity signatures from others.

contracts/sophia.md
```diff
@@ -814,9 +840,16 @@ private public rec stateful switch true type record datatype
...
-- `Hash = #[0-9A-Fa-f]+` hash literal
+- `Bytes = #[0-9A-Fa-f]+` byte array literal
 - `String` string literal enclosed in `"` with escape character `\`
 - `Char` character literal enclosed in `'` with escape character `\`
+- `AccountAddress` base58-encoded 32 byte account pubkey with `ak_` prefix
+- `ContractAddress` base58-encoded 32 byte contract address with `ct_` prefix
+- `OracleAddress` base58-encoded 32 byte oracle address with `ok_` prefix
+- `OracleQueryId` base58-encoded 32 byte oracle query id with `oq_` prefix
+
+See the [identifier encoding scheme](./node/api/api_encoding.md) for the
```
lucafavatella: Broken link:
```suggestion
See the [identifier encoding scheme](/node/api/api_encoding.md) for the
```

I see this page names base58 though meaning Base58Check, and that is clarified in this link. Ok.

contracts/sophia.md
```diff
@@ -241,22 +241,47 @@ Sophia has the following types:
...
 | map        | An immutable key value store with dynamic mapping of keys of one type to values of one type | ```type accounts = map(string, address)```
 | option('a) | An optional value either None or Some('a) | ```Some(42)```
-| state        | A record of blockstate key, value pairs  |
-| transactions | An append only list of blockchain transactions |
-| events       | An append only list of blockchain events (or log entries) |
-| signature    | A signature - 64 bytes |
-| Chain.ttl    | Time-to-live (fixed height or relative to current block) | ```FixedTTL(1050)``` ```RelativeTTL(50)```
+| state      | A user defined type holding the contract state | ```record state = { owner: address, magic_key: bytes(4) }```
+| event      | An append only list of blockchain events (or log entries) | ```datatype event = EventX(indexed int, string)```
+| hash       | A 32-byte hash - equivalent to `bytes(32)` |
```
lucafavatella: I see from [here](https://github.com/aeternity/aesophia/compare/faa0ef97725f82dedee9e026c88c428494ce9c2c...386419f11256407c85ae96c6afcd122fe4d4506f#diff-187d952da8f4fdac5d1ad407bb58fd91R382). Ok.

contracts/sophia.md
```diff
@@ -241,22 +241,47 @@ Sophia has the following types:
...
+| signature  | A signature - equivalent to `bytes(64)` |
```
lucafavatella: I see from [here](https://github.com/aeternity/aesophia/compare/faa0ef97725f82dedee9e026c88c428494ce9c2c...386419f11256407c85ae96c6afcd122fe4d4506f#diff-187d952da8f4fdac5d1ad407bb58fd91R383). Ok.

contracts/sophia.md
```diff
@@ -814,9 +840,16 @@ private public rec stateful switch true type record datatype
 - `QCon = (Con\.)+Con` qualified constructor
 - `TVar = 'Id` type variable (e.g `'a`, `'b`)
 - `Int = [0-9]+|0x[0-9A-Fa-f]+` integer literal
-- `Hash = #[0-9A-Fa-f]+` hash literal
+- `Bytes = #[0-9A-Fa-f]+` byte array literal
```
lucafavatella: Type `bytes(n)` is defined above as "A byte array with `n` bytes". This grammar allows for odd number of hexadecimal digits.

Is an odd number of hexadecimal legal?
If yes, is it equivalent to same literal prepended by one zero?

lucafavatella: Is any canonical representation enforced by compiler e.g. no preceding zeros?

hanssv: An odd number of hex-digits means there is an implicit extra `0`...

Nothing is enforced by the compiler, leading zeros have a meaning.

PR #343, opened 2019-04-29
Describe off-chain protocol pinned env

velzevur:
Corresponding changes for aeternity/aeternity#2369


channels/OFF-CHAIN.md
```diff
@@ -416,13 +416,17 @@ Message code: 5
 In order to open a channel on chain, parties need to cooperate and co-sign the
 `channel_create` transaction. The `funding_created` message is used by the
 initiator to send an initial state - a `channel_create_tx` object, signed
-by the Initiator.
+by the Initiator. The `block_hash` specifies the on-chain environment one
```
hanssv: This sentence doesn't read well...

The block_hash field specifies which on-chain environment was used to prepare... or something similar?!

PR #346, opened 2019-05-07
Update bitcoin-ng.md

fedecaccia:
correcting semantic error


sennui:
@fedecaccia thank you for catching the error

PR #347, opened 2019-05-09
Add address check primitives

hanssv:
Added `Address.is_contract`, `Address.is_oracle`, `Oracle.check` and `Oracle.check_query`.

And a note on `Contract.creator`...

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -519,7 +519,9 @@ The following builtin functions are defined on integers:
 The following builtin functions are defined on addresses:
 
 ```
-  Address.to_str(a : address) : string  // Base58 encoded string
+  Address.to_str(a : address) : string    // Base58 encoded string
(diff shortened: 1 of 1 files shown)


contracts/sophia.md
```diff
@@ -798,7 +822,7 @@ The block-chain environment available to a contract is defined in three name spa
 `Contract`, `Call`, and `Chain`:
 
 - `Contract.creator` is the address of the entity that signed the contract creation
-  transaction.
+  transaction. (Available from Fortuna release.)
```
lucafavatella: Contract creation transaction only? Or also generalized accounts attach transaction?

hanssv: The attach is considered a contract creation in this aspect.

hanssv:
contracts/aevm.md is horribly outdated, in need of a touch-up...

hanssv:
> Where would I run this?

This the protocol repository it only has documentation...

PR #345, opened 2019-05-05
Details of rewards including rewards split

sennui:

PR #350, opened 2019-05-13
Details of rewards including rewards split

lucafavatella:
Alternative to #345

The idea is to have the text ready so we can merge when/if needed.


thepiwo:
https://github.com/aeternity/protocol/pull/355 builds on top of this

thepiwo:
image
 voting results

PR #352, opened 2019-05-14
Document stateful checks

UlfNorell:
cc https://github.com/aeternity/aesophia/pull/69

### DIFF
--- a/contracts/sophia.md
+++ b/contracts/sophia.md
@@ -732,36 +732,36 @@ Example for an oracle answering questions of type `string` with answers of type 
 ```
 contract Oracles =
 
-  function registerOracle(acct : address,
-                          sign : signature,   // Signed oracle address + contract address
-                          qfee : int,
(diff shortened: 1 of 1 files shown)


contracts/sophia.md
```diff
@@ -184,11 +184,45 @@ state associated with each contract instance.
...
+    put(s)
+```
+
+Without the `stateful` annotation the compiler does not allow the call to
+`put`. A `stateful` annotation is required to
+
+* Use a stateful primitive function. These are
+  - `put`
+  - `Chain.spend`
+  - `Oracle.register`
```
lucafavatella: Could you please clarify https://github.com/aeternity/protocol/blame/89a8d7511248248b0e42bab3fe2970703fe49b87/contracts/sophia.md#L735-L739

I would expect stateful there? Could you please check the rest of that example?

UlfNorell: Yeah, those are missing some `stateful` annotations. I'll update.

PR #353, opened 2019-05-14
Misc fixes - byte array, minumum fee, etc

hanssv:
https://www.pivotaltracker.com/story/show/165993503 and https://www.pivotaltracker.com/story/show/165853843

### DIFF
--- a/consensus/consensus.md
+++ b/consensus/consensus.md
@@ -227,7 +227,7 @@ Each (on-chain) transaction has the following fields:
 
 *Note*: There is also a node configurable minimum gas price - this is the
 minimum gas price a node accepts and is not under consensus. This value is
-normally higher than the minimal gas price as dictated by consensus. For
+higher than or equal to the minimal gas price as dictated by consensus. For
 contract transactions (Create, Call, GAAttach, GAMeta), the minumum applies to
 both the `Fee` and the `GasPrice`.
 
--- a/consensus/consensus.md
+++ b/consensus/consensus.md
@@ -284,7 +284,7 @@ The gas of a transaction is the sum of:
 | Contract create        | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(ContractCreateTx) * GasPerByte`. It also includes gas for contract execution. |
 | Contract call          | `30 * BaseGas` | Proportional to the byte size of the transaction, specifically: `byte_size(ContractCallTx) * GasPerByte`. It also includes gas for contract execution. |
 | GA Attach              | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAAttachTx) * GasPerByte`. It also includes gas for execution of the init function. |
-| GA Meta                | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAMetaTx) * GasPerByte`. It also includes gas for execution of the authentication function + gas for the wrapped transaction. |
+| GA Meta                | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAMetaTx) * GasPerByte`. It also includes gas for execution of the authentication function + recursively gas corresponding to the wrapped transaction(s). |
 
 `BaseGas` is 15 000.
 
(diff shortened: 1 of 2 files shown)


consensus/consensus.md
```diff
@@ -225,6 +225,12 @@ Each (on-chain) transaction has the following fields:
   `1` (*10^-18) aeons.)
 * Time to live (TTL).
 
+*Note*: There is also a node configurable minimum gas price - this is the
+minimum gas price a node accepts and is not under consensus. This value is
+normally higher than the minimal gas price as dictated by consensus. For
```
velzevur:
```suggestion
higher or equal to the minimal gas price as dictated by consensus. For
```

node/api/api_encoding.md
```diff
@@ -32,6 +32,7 @@ tag.
 | Prefix  | Encoding | Description |
 | ---     | ---      | --- |
 | ak      | base58   | Account pubkey |
+| ba      | base64   | Byte array |
```
lucafavatella: This brought in by https://github.com/aeternity/aeternity/pull/2415/files#diff-31d7a50c99c265ca2793c20961b60979R68
Ok.

consensus/consensus.md
```diff
@@ -105,18 +105,18 @@ The peer-to-peer network is not part of the consensus and will be described in a
...
 
-#### Aeons
+#### Aeternity tokens
 
-Aeon is the main coin used in the Æternity system. We use the following denominations:
+We use the following denominations:
 
 ```
-1
+1                          | aetto
```
lucafavatella: From https://en.wikipedia.org/wiki/Atto-
Ok.

consensus/consensus.md
```diff
@@ -277,6 +283,8 @@ The gas of a transaction is the sum of:
 | Channel offchain       |            0   |     0 |
 | Contract create        | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(ContractCreateTx) * GasPerByte`. It also includes gas for contract execution. |
 | Contract call          | `30 * BaseGas` | Proportional to the byte size of the transaction, specifically: `byte_size(ContractCallTx) * GasPerByte`. It also includes gas for contract execution. |
+| GA Attach              | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAAttachTx) * GasPerByte`. It also includes gas for execution of the init function. |
+| GA Meta                | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAMetaTx) * GasPerByte`. It also includes gas for execution of the authentication function + gas for the wrapped transaction. |
```
lucafavatella: Considering [`aetx:gas_limit`](https://github.com/aeternity/aeternity/blob/be71889c61c107cd848e97d3c58623ca0e93581b/apps/aetx/src/aetx.erl#L257-L258) and `aega_meta_tx:gas_limit`, I do not see how the base gas of the inner tx is not accounted for this gas (i.e. equivalent gas of tx for micro block gas limit).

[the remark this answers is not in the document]
hanssv: I think that this is taken care of by https://github.com/aeternity/aeternity/pull/2362 when it comes to actual charging. But the `gas_limit` will indeed make use be a bit pessimistic when it comes to picking what candidate transactions may fit in a microblock. This could be made more precise but with some additional complexity.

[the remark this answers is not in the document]
hanssv: It should read `transactions` - it is obviously for transactions that have two signing parties. Also, while you may not add superfluous "signatures" (i.e. wrapping layers) it is actually allowed for the same account to sign (i.e wrap) multiple times though I see no real use case for it.

For channels there are several places where multiple wrapping may happen.

hanssv: "gas for the wrapped transactions" is not very precise, it would and should include also base gas...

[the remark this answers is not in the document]
lucafavatella: Tracked https://www.pivotaltracker.com/story/show/166018387

consensus/consensus.md
```diff
@@ -277,6 +283,8 @@ The gas of a transaction is the sum of:
 | Channel offchain       |            0   |     0 |
 | Contract create        | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(ContractCreateTx) * GasPerByte`. It also includes gas for contract execution. |
 | Contract call          | `30 * BaseGas` | Proportional to the byte size of the transaction, specifically: `byte_size(ContractCallTx) * GasPerByte`. It also includes gas for contract execution. |
+| GA Attach              | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAAttachTx) * GasPerByte`. It also includes gas for execution of the init function. |
+| GA Meta                | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAMetaTx) * GasPerByte`. It also includes gas for execution of the authentication function + recursively gas corresponding to the wrapped transaction(s). |
```
lucafavatella:
```suggestion
| GA Meta                | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAMetaTx) * GasPerByte`. It also includes gas for execution of the authentication function + recursively gas corresponding to the wrapped transaction(s) (excluding the byte size portion - in order not to account for the size of wrapped transactions multiple times). |
```

PR #358, opened 2019-05-20
Add Sophia Explained

hanssv:
This is planned to be a place to add more in-depth discussion on Sophia topics...

### DIFF
--- a/contracts/sophia_explained.md
+++ b/contracts/sophia_explained.md
@@ -88,7 +88,24 @@ of the event constructor name, i.e. `Keccak("TheFirstEvent")` and
   157,0,241,233,150,150,117,1,154,28,43,125,65,8,...>>
 ```
 
-The implicit value goes first in the `topics` array.
+The implicit value goes first in the `topics` array. The topics are (currently,
+this might change in the future) presented as 256-bit unsigned integers - i.e.
+`1875564187002476023854543820981509249331367502514562901623081521315128929137`
+in the second event in the example corresponds to `<<"ct_2pvLLjPfjq...">>`. A
+boolean argument would come out as `0` (= false) or `1` (= true).
+
+### Argument order
+
+It is only possible to have one (1) `string` parameter in the event, but it can
+be placed in any position (and its value will end up in the `data` field), i.e.
+```
+AnotherEvent(string, indexed address)
+
+...
+
+Chain.event(AnotherEvent("This is not indexed", Contract.address))
+```
+would yield exactly the same result in the example above!
 
 ### Event indexing
 
--- a/contracts/sophia_explained.md
+++ b/contracts/sophia_explained.md
@@ -1,0 +1,98 @@
+[back](./contracts.md)
+# Sophia explained
+
+The [Sophia](./sophia.md) language is described in [The Sophia
+Language](./sophia.md). However, some concepts could need a more in depth
+description and some accompanying examples to be easier to grasp. We aim to
+collect such in-depth explanations on this page.
+
+## Table of Contents
+- [Events](#events)
+
+## Events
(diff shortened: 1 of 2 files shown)


contracts/sophia_explained.md
```diff
@@ -0,0 +1,99 @@
+[back](./contracts.md)
```
dincho: double back

hanssv: Copy pasted from `contracts/sophia.md` :joy:

Will fix.

hanssv: Fixed.

hanssv:
> I would like just to make sure if I got it right:
> 
>     * the `data` field is the latest event parameter, which is always string and always non-indexed (what non-indexed means ?)
No, the position is arbitrary it can be in any position, but only one and non-indexed. The `indexed` keyword was a mistake I think,  it made sense in an earlier variant of the design. 
> 
>     * the event parameters are called topics ? Except the last one, non-indexed, string parameter, which is called data?
Yes
> 
>     * the hash of event name/constructor is always the first topic
Yes 
>     * I assume there cannot be same constructor/different parameters (signature) ?
No, this is built-into Sophia itself, a datatype can not have overlapping constructors.
> 
> I find the terminology and fields just a little bit confusing, but otherwise this docs (assuming above is true) explains it well. 👍

dincho:
> No, the position is arbitrary it can be in any position, but only one and non-indexed. The `indexed` keyword was a mistake I think, it made sense in an earlier variant of the design.

OK, so the data is then defined by the non-indexed (without indexed keyword) string param ?

Then if I wrote it as:
```
AnotherEvent(string, indexed address)
```

And then:

```
	  public function emit_events() : () =
	    Chain.event(TheFirstEvent(42))
	    Chain.event(AnotherEvent("This is not indexed", Contract.address))
```

I should get the exactly the same output ?

hanssv:
> > No, the position is arbitrary it can be in any position, but only one and non-indexed. The `indexed` keyword was a mistake I think, it made sense in an earlier variant of the design.
> 
> OK, so the data is then defined by the non-indexed (without indexed keyword) string param ?
> 
> Then if I wrote it as:
> 
> ```
> AnotherEvent(string, indexed address)
> ```
> 
> And then:
> 
> ```
> 	  public function emit_events() : () =
> 	    Chain.event(TheFirstEvent(42))
> 	    Chain.event(AnotherEvent("This is not indexed", Contract.address))
> ```
> 
> I should get the exactly the same output ?

Yes, exactly!

contracts/sophia_explained.md
```diff
@@ -0,0 +1,111 @@
...
+                          1875564187002476023854543820981509249331367502514562901623081521315128929137]}],
+  <<"return_type">> => <<"ok">>,
+  <<"return_value">> => <<"cb_AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAArMtts">>}
+```
+
+First look at the `data` field. For the first event it is `<<"cb_Xfbg4g==">>`
+which decodes to `<<>>`, i.e. no data (and that is expected since the event did
+not have any extra data.). For the second event we see
+`<<"cb_VGhpcyBpcyBub3QgaW5kZXhlZK+w140=">>` which decodes to `<<"This is not
+indexed">>`. For the `topics` field we see that there is actually two values
```
maptuhec: I think that the first value of the `topics` field can be explained a bit more. For the first example event the `indexed int` value is presented as a plain integer (42), but for the second event the `indexed address` value is, I assume, hash of the address. The logic behind this is not explained, what would be the value if the parameter is a `boolean`?

PR #364, opened 2019-06-04
Describe the ping-pong mechanism

velzevur:
Complementary changes for aeternity/aeternity#2484

### DIFF
--- a/node/api/channels_api_usage.md
+++ b/node/api/channels_api_usage.md
@@ -78,6 +78,8 @@ but are not necessarily part of the channel's life cycle.
 
 * [Cleaning local contract calls](#pruning-contract-calls)
 
+* [Connection keep alive](#connection-keep-alive)
+
 Only steps 1 and 4 require chain interactions, step 2 and 3 are off-chain.
 
 ### HTTP requests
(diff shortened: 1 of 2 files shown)


node/api/channels_api_usage.md
```diff
@@ -2056,3 +2058,51 @@ The transaction has the following structure:
...
+they have any control over those frames. They are to be handled by browsers
+and this could introduce some undesired incompatibilities.
+
+In order to provide best user experience, we've kept the functionality of the
+node to respond with a `pong` control frame to every `ping` received as well
+as enhanced the State Channel's WebSocket API with the option of sending data
+frames that act like the corresponding control frames. Depending on the
+environment to be ran, the client can use either approach.
+
+If no frames had been received for 1 minute, the node will consider the
```
lucafavatella: I did not look in details where this 1 minute is set in the code. It seems too detailed here, and shall rather be user configurable, with default mentioned in the user configuration schema.

velzevur: This comes from the `cowboy` lib and it is out of the box feature. Please check section `Keeping the connection alive` in [the docs](https://ninenines.eu/docs/en/cowboy/2.1/guide/ws_handlers/). It could as well be connection specific config but it is not exposed to State Channels' clients. I am not really eager exposing it as it could easily lead to clients leaking WebSocket connections on the node.

node/api/channels_api_usage.md
```diff
@@ -2056,3 +2058,51 @@ The transaction has the following structure:
...
+Thus the WebSocket protocol defines special control frames to be used: sending
+`ping` and `pong`. Clients that are run in a browser have no access nor do
+they have any control over those frames. They are to be handled by browsers
+and this could introduce some undesired incompatibilities.
+
+In order to provide best user experience, we've kept the functionality of the
+node to respond with a `pong` control frame to every `ping` received as well
+as enhanced the State Channel's WebSocket API with the option of sending data
+frames that act like the corresponding control frames. Depending on the
+environment to be ran, the client can use either approach.
```
uwiger: ran -> used | run

node/api/channels_api_usage.md
uwiger: had been -> have been

PR #370, opened 2019-07-01
fix mistake in example

mradkov:
- Fixing a mistake in constructing record via function example

The passed parameter was not given a type and the function was not returning any value which it should as this was the example aiming to do.


hanssv:
Adding type annotations doesn't change what the function returns 🤔

mradkov:
Shouldn’t there be an explicit return type?

hanssv:
Not necessarily, the types are inferred, if the compiler can't figure it out it will complain.

mradkov:
Feel free to close this pr then. :)

PR #372, opened 2019-07-03
Update protocol with possibility to sign just the Tx Hash

hanssv:


consensus/consensus.md
```diff
@@ -71,22 +71,28 @@ account to smart contracts.
...
 Signature = sign(NetworkId + SerializedObject)
+or Signature = sign(NetworkId + Blake2b(SerializedObject))
 ```
 
 Prefix defaults to ``ae_mainnet`` (``binary()``) and it is configurable via node config.
 The prefix is not part of a serialized transaction. We add it only for signature.
 Consider changing the id in case of forking the blockchain network,
 in order to lower danger of replay attacks.
 
+**Note:** Post Lima hard-fork it is allowed to sign either the full serialized object, or
```
lucafavatella: The definition of "Lima" is introduced in https://github.com/aeternity/protocol/pull/391

It would be great if you could have a look to that one too.

PR #374, opened 2019-07-12
Define WebSocket opening errors

velzevur:
Documentation regardin /aeternity/aeternity#2560

### DIFF
--- a/node/api/channels_api_usage.md
+++ b/node/api/channels_api_usage.md
@@ -58,6 +58,8 @@ expected. The flow is the following:
 There are a some WebSocket events that can occur while the connection is open
 but are not necessarily part of the channel's life cycle.
 
+* [Open error](#open-error)
+
 * [Update error](#update-error)
 
 * [Update conflict](#update-conflict)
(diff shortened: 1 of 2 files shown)


node/api/channels_api_usage.md
```diff
@@ -1489,6 +1491,77 @@ is finally closed, and the channel object removed:
...
+      "request":{
+      }
+   },
+   "id":null,
+   "jsonrpc":"2.0",
+   "version":1
+}
+```
+
+Examples for this would be either opening amount being bellow the threshold
```
lucafavatella:
```suggestion
Examples for this would be either opening amount being below the threshold
```

node/api/channels_api_usage.md
```diff
@@ -1489,6 +1491,77 @@ is finally closed, and the channel object removed:
...
+defined by `channel_reserve`, or any of `channel_reserve`, `push_amount` or
```
lucafavatella: It would be nice if the error mentioned which field is wrong.

PR #375, opened 2019-07-16
Fix markup re GAs

lucafavatella:


generalized_accounts/generalized_accounts.md
```diff
@@ -340,5 +340,5 @@ Some mechanisms, that we had in mind, as motivation for GA:
 - multi-sig (e.g. phone + hardware wallet, or multiple people)
 - using different signature algorithms (i.e. other than EdDSA)
 
-GAs are further explained and an example (EdDSA) is provided in (Generalized
-accounts explained)[ga_explained.md].
+GAs are further explained and an example (EdDSA) is provided in [Generalized
```
lucafavatella: This mentions EdDSA while the linked doc mentions EcDSA.
```suggestion
GAs are further explained and an example is provided in [Generalized
```

PR #378, opened 2019-07-16
Add info on SC port multiplexing support

uwiger:
Documentation for [PT #167194534](https://www.pivotaltracker.com/story/show/167194534) and [aeternity PR #2572](https://github.com/aeternity/aeternity/pull/2572)

### DIFF
--- a/node/api/channels_api_usage.md
+++ b/node/api/channels_api_usage.md
@@ -109,7 +109,11 @@ and another for optional timeouts.
 
   `responder`'s port and host pair must be reachable from `initiator` network
   so unless participants are part of a LAN, they should be exposed to the
-  internet as described [here](../../node/api/README.md).
+  internet as described [here](../../node/api/README.md). It is possible to use the
+  same port number for different responder pubkeys and multiple simultaneous
+  responders. If the `responder` sets `initiator_id` to `"any"`, the responder will
+  accept a connection request from any initiator, and fetch the proper `initiator_id`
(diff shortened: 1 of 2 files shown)


node/api/channels_api_usage.md
```diff
@@ -109,7 +109,11 @@ and another for optional timeouts.
 
   `responder`'s port and host pair must be reachable from `initiator` network
   so unless participants are part of a LAN, they should be exposed to the
-  internet as described [here](../../node/api/README.md).
+  internet as described [here](../../node/api/README.md). It is possible to use the
+  same port number for different responder pubkeys and multiple simultaneous
+  responders. If the `responder` sets `initiator_id` to `"any"`, the responder will
```
tolbrino: I don't think its a clean API design to overload the meaning of parameters with multiple possible types. I'd prefer to use a separate parameter to indicate that multiplexing is allowed.

Also where is the `"any"` actually evaluated in the code?

lucafavatella: Also "any" does not fit description above "initiator's public key".

lucafavatella: > Also where is the `"any"` actually evaluated in the code?

It seems [here](https://github.com/aeternity/aeternity/blob/91e4fe12721e108a477e116d7f49ba40fc3d9c1a/apps/aehttp/src/sc_ws_handler.erl#L309).

PR #382, opened 2019-07-25
VSC yellow paper

velzevur:
For some time now @uwiger and I had been describing our vision for Virtual State Channel.
This PR is the current view of it, once we all agree upon it, we will share it to the forum where people can discuss it even further.

The goal is defining the on-chain protocol.

### DIFF
--- a/channels/VSC.md
+++ b/channels/VSC.md
@@ -412,19 +412,63 @@ closing sequence in their on-chain channel with the intermediary. It results in
 modifying the channel object there with an appropriate on-chain height to be
 reached in order for the channel to be solo-closed. Meanwhile the intermediary
 does a corresponding operation with the other party, giving it time to provide a
-newer off-chain state, using a on-chain forced progress if necessary.
+newer off-chain state, using a on-chain virtual forced progress if necessary.
 
 It is not possible to reduce the amount of the other party in the VSC object
 during a unilateral update, i.e. the 'requester' cannot reduce the
 `intermediary_amount` and the 'intermediary' cannot reduce the
 `requester_amount`. This is ultimately checked on-chain, if needed.
 
+##### Virtual State Channel Force Progress
+
+In order to improve privacy and not to provide the whole SC state, we define a
+new transaction type for Virtual State Channel Force Progressing. Instead of
+providing the whole SC state tree, only a Proof of Inclusion containing the
+relevant object is provided. Depending on the specific progress would include
+the channel object and optionally accounts to which tokens can be withdrawn
+to.
+
+The transaction is called `channel_vsc_force_progress` and it consists of:
+* `payload` that defines the `round` and the `state_hash` the progress is base
+  on. This could be:
+    * the latest on-chain persisted one: in this case the on-chain stored
+      `round` and `state_hash` are used. The `payload` is empty
+    * off-chain transaction with a greater `round` to the one currently
+      persisted on-chain for this SC. The `payload` is a serializated
+      off-chain transaction and its `round` and `state_hash` are used for
+      defining the SC state that the progress is being based upon.
+* `poi` is a Proof of Inclusion that provides enough information for the
+  forced progress to be applied. This would include the VSC object but
+  possibly accounts to which tockens are withdrawn to might be needed as well.
+* `update` is a VSC force progress update. This could be:
+  * a co-authenticated VSC withdraw
+  * a co-authenticated VSC snapshot
+  * a co-authenticated VSC settle
+  * a co-authenticated VSC mutual close
+  * unilaterally authenticated dispute. Those also provide a context of the
+    execution: a `payload` that is either empty or a co-authenticated VSC
+    off-chain state, VSC poi or trees, next VSC `state_hash` and next VSC
+    `round`. Update types for disputes supported are:
+    * solo close
+    * slash
+    * force progress
+    * virtual force progress
+* `state_hash` - the root of the SC state tree after the `update` had been
+  applied on top of the `poi`
+* `round`: SC's next round
+
+Note that this allows force progressing of both contracts in VSC and force
+progressing a VSC inside a VSC.
+
+From now on we will use force progress in the context of VSC meaning a Virtual
+State Channel Force progress transaction - `channel_vsc_force_progress`.
+
 ##### Example
 
 Alice wants to close her Virtual State Channel with Bob, but Bob refuses. Alice
 produces a solo update in her on-chain State Channel with Ingrid initiating a
 solo-closing sequence of the Virtual State Channel. If Ingrid refuses this,
-Alice can force it on-chain.
+Alice can force it on-chain using `channel_vsc_force_progress.
 
 Ingrid takes this update provided by Alice and applies it to her on-chain State
 Channel with Bob. If Bob refuses, this can be force-progressed as well.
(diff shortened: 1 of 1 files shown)


uwiger:
I will withhold my review, since I was a co-author of the document. :)

velzevur:
> The proposal reads more like a blog post and less like a specification / technical design. If that is not the intention it should be reworked to be more formal and structured.

Yes, as intention is to be shared to the community so it can be discussed there. Later on we will have a better more strict and technical documentation.

Regarding the approach - the goal is to have a solid shared understanding regarding on-chain dispute mechanisms, so we can bake the in the on-chain protocol. The details of tthe off-chain protocol are not yet cleared out and are outside of the scope of this document.

channels/VSC.md
```diff
@@ -0,0 +1,739 @@
...
+  |  off-chain  |                     |  off-chain  |
+  |    state    |                     |    state    |
+  |    trees    |                     |    trees    |
+  |             |                     |             |
+  +-------------+                     +-------------+
+```
+
+Each channel has an ever changing state. It is represented by a set of MPTrees
+that hold accounts, contracts and contract calls. This set of MPtrees can be
+treated as one big MPTree that has a root hash. This is what we call a
```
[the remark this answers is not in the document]
velzevur: yes

velzevur:
> 1\. There should be a TX dedicated for force progressing VSC operations - more info in devchat.
> 2\. The only case in which the VSC state should be exposed is when on-chain forcing a force_progress_tx inside the VSC - more info in devchat. Additionally the AI, IB offchain state can remain mostly private.

This is true, I will amend the file with it.

---


> 3\. Channel creation should sometimes be force progress-able - For instance Alice sends to Ivory a channel create TX and then Ivory applies the TX to the AI channel but not to the IB channel - In this case Bob or Alice should be able to perform a VSC force progress operation.

In this case Ivory will have already commited some tokens to the VSC in Alice-Ivory SC, so Ivory has the financial incentive to open the same VSC in Ivory-Bob. If Bob rejects it - Ivory shall close the VSC in Alice-Ivory VSC. Until Bob gets the virtual channel creation in Bob-Ivory SC, then Bob can not proceed using the VSC in a safe manner.

---

> 4\. Onchain VSC force progress operations could in some cases move the channel onchain(If Ivory cheated once then she may do it again). The proposed channel_vsc_force_progress_tx contains enough information to move the channel onchain - after a move was done we need to wait for disputes comming from Ivory as the channel_vsc_force_progress_tx may not be based on the latest state.

I find this a cool optimization but it is debatable if it is needed for the initial version

---

> 5\. Sometimes we may want to change the intermediary - this can be done if the parties cooperate. I'm not sure if this will be useful.

Indeed useful and will make the netwok much more dynamic and resilient. Since it is outside of the scope of the on-chain protocol, it is outside of the scope of the document.

---

> 6\. I'm not sure how disputes should be handled in the multi-intermediary VSC, If we stick to the VSC in VSC model then I see how can we guarantee security.

I think it would be the same as everyone can take care just for themselves.

---

> 7\. The off chain FSM should check if an VSC can be established - if so then while opening a channel the client should be able to chose between creating an onchain SC or a VSC - this would increase usability drastically.

FSM is out of the scope of this document.

channels/VSC.md
```diff
@@ -0,0 +1,783 @@
...
+}
+```
+
+Note the amounts as we will revisit those with regard of Virtual State Channels
+
+#### Disputes
+
+The general assumption is if the protocol is clean and easy to use, this would
+provide incentive for participants to follow it. Cheating would be meaningless
+as it would yeld no benefit for the cheating party. That is a strong argument in
```
AleksandarFilipov: yield (yeld), typo

channels/VSC.md
```diff
@@ -0,0 +1,783 @@
...
+         "initiator_id": "ak_...", // the address of Alice
+         "responder_id": "ak_...", // the address of Bob
+         "initiator_amount": 10, // Alice balance
+         "responder_amount": 12, // Bob balance
+         "channel_amount": 22 // the sum of the two amounts
+      } ...
+}
+```
+
+We could have an implicit logic that in the context of this particulat State
```
AleksandarFilipov: typo: particulat

channels/VSC.md
```diff
@@ -0,0 +1,783 @@
...
+according to her authentication method used in her on-chain State Channel with
+Ingrid. She then gives it to Bob to co-authenticate. Bob does so according to
+his authentication method in his on-chain State Channel with Ingrid. He then
+returns the co-authenticated transaction to Alice and the update is finished.
+
+It is worth noting that this closly resembles the on-chain State Channel
+protocol. Also it is important to stress that none of the off-chain updates
+reach Ingrid.
+
+#### Depositing more tokens to a Virtual State Channel/withdrawing from a
```
AleksandarFilipov:
```suggestion
#### Depositing more tokens to a Virtual State Channel/withdrawing from a Virtual State Channel
```

channels/VSC.md
```diff
@@ -0,0 +1,783 @@
...
+Virtual State Channel
```
AleksandarFilipov:
```suggestion
```

AleksandarFilipov:
1. Upon VSC creation Ingrid must have sufficient amount of AE:s, the same goes when Alice/Bob does a deposit (this would strain Ingrid). 
2. Upon creation of a VSC Alice and Bob are re-allocation a portion of his/her AE to the new VSC. These were previously allocated on SC Alice-Ingrid and Bob-Ingrid. This would present some kind of risk to Ingrid?

velzevur:
> 1\. Upon VSC creation Ingrid must have sufficient amount of AE:s, the same goes when Alice/Bob does a deposit (this would strain Ingrid).

Yes, this makes sure the channel is not having more tokens off-chain than on-chain. Otherwise it would be undisputable on-chain.

> 2\. Upon creation of a VSC Alice and Bob are re-allocation a portion of his/her AE to the new VSC. These were previously allocated on SC Alice-Ingrid and Bob-Ingrid. This would present some kind of risk to Ingrid?

The goal is Ingrid to be safe in any case, including the one where Alice and Bob are not cooperating and refuse closing the channel. Then Ingrid can initate a closing on-chain and let them provide newer states than the opening one.

channels/VSC.md
```diff
@@ -0,0 +1,782 @@
...
+##### Alice refuses to cooperate with Ingrid
+
+If Ingrid has a newer state than the one Alice had provided, or Ingrid has a
+state produced by a valid force progress by Bob, Alice might decide not to
+cooperate. In this case Ingrid simply brings the issue on-chain using a forced
+progress.
+
+##### Alice produces a force progress based on an old state
+
+Ingrid receives a forced progress that seemsreasonable because the state it is
```
AleksandarFilipov:
```suggestion
Ingrid receives a forced progress that seems reasonable because the state it is
```

channels/VSC.md
```diff
@@ -0,0 +1,782 @@
...
+based on a is the latest she had seen so far. Ingrid accepts it as valid and
```
AleksandarFilipov:
```suggestion
based on is the latest she had seen so far. Ingrid accepts it as valid and
```

AleksandarFilipov:
1. Closing the VSC would mean that we the go back to the state before opening. Difference being that the AE:s between Bob and Ingrid are distributed differently. The same same does for the AE:s between Alice and Ingrid. Critical is that Ingrids AE:s in the two SC equal the same amount as before the VSC.
So if Alice opens a VSC (with bob over Ingrid) this would mean that there is a withdrawal from the SC (between Alice and Ingrid). How is that reflected in Alice-Ingrid SC?

2. Assume that Alice-Ingrid SC could operate at the same time as Alice-Bob. What kind of complexity does this introduce, given that the allocation changes for the carrier SC. Same reasoning goes for VSC upon other VSC.

3. Fuzzy one, https://github.com/aeternity/protocol/pull/382/files#diff-27f917ebdc0662bc956375a713a87123R386 It would make sense to try and distigish the term _on-chain_ when its acutally referrs to Ingrid or node communication.

channels/VSC.md
```diff
@@ -412,19 +412,63 @@ closing sequence in their on-chain channel with the intermediary. It results in
...
+  on. This could be:
+    * the latest on-chain persisted one: in this case the on-chain stored
+      `round` and `state_hash` are used. The `payload` is empty
+    * off-chain transaction with a greater `round` to the one currently
+      persisted on-chain for this SC. The `payload` is a serializated
+      off-chain transaction and its `round` and `state_hash` are used for
+      defining the SC state that the progress is being based upon.
+* `poi` is a Proof of Inclusion that provides enough information for the
+  forced progress to be applied. This would include the VSC object but
+  possibly accounts to which tockens are withdrawn to might be needed as well.
```
uwiger:
```suggestion
  possibly accounts to which tokens are withdrawn might be needed as well.
```

channels/VSC.md
```diff
@@ -412,19 +412,63 @@ closing sequence in their on-chain channel with the intermediary. It results in
...
+* `update` is a VSC force progress update. This could be:
+  * a co-authenticated VSC withdraw
+  * a co-authenticated VSC snapshot
+  * a co-authenticated VSC settle
+  * a co-authenticated VSC mutual close
+  * unilaterally authenticated dispute. Those also provide a context of the
```
uwiger: _a_ unilaterally ...
context _for_ ...

channels/VSC.md
```diff
@@ -412,19 +412,63 @@ closing sequence in their on-chain channel with the intermediary. It results in
...
+    `round`. Update types for disputes supported are:
+    * solo close
+    * slash
+    * force progress
+    * virtual force progress
+* `state_hash` - the root of the SC state tree after the `update` had been
+  applied on top of the `poi`
+* `round`: SC's next round
+
+Note that this allows force progressing of both contracts in VSC and force
```
uwiger: force-progressing

channels/VSC.md
```diff
@@ -412,19 +412,63 @@ closing sequence in their on-chain channel with the intermediary. It results in
...
+progressing a VSC inside a VSC.
+
+From now on we will use force progress in the context of VSC meaning a Virtual
+State Channel Force progress transaction - `channel_vsc_force_progress`.
```
uwiger: Force Progress

channels/VSC.md
```diff
@@ -412,19 +412,63 @@ closing sequence in their on-chain channel with the intermediary. It results in
...
+
 ##### Example
 
 Alice wants to close her Virtual State Channel with Bob, but Bob refuses. Alice
 produces a solo update in her on-chain State Channel with Ingrid initiating a
 solo-closing sequence of the Virtual State Channel. If Ingrid refuses this,
-Alice can force it on-chain.
+Alice can force it on-chain using `channel_vsc_force_progress.
```
uwiger: missing ending backtick.

channels/VSC.md
```diff
@@ -443,38 +487,38 @@ manner.
...
-Alice safe. In order for Ingrid to be safe, she can propose to Bob to execute
-the same update Alice had already provided her, this time in Ingrid-Bob's State
-Channel. If Bob still refuses, Ingrid can produce an actual
-`channel_force_progress` transaction that she posts on-chain in order for her to
-enforce the contract execution in the Virtual State Channel, thus producing the
-new Alice-Bob Virtual State Channel on-chain and at the same time producing the
-next Ingrid-Bob off-chain state. The newly produced Ingrid-Bob state has in its
-state the VSC's.
+refuses to co-authenticate the next state, Alice can produce an off-chain
+`channel_vsc_force_progress` which is similar to force progress and bring it
```
uwiger: ..., which is similar to force progress, ... (commas)

PR #385, opened 2019-07-26
PT-166642114 Miner signalled consensus upgrade

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/166642114

TODO:
* [x] Validate this PR by design in the reference implementation.
  - See draft design at https://github.com/aeternity/aeternity/pull/2613
* [x] Review/confirm/refine/revise the value in the key block header field `info` being part of the validation rules of the new consensus protocol on the "signalling" block interval.

### DIFF
--- a/consensus/consensus.md
+++ b/consensus/consensus.md
@@ -170,6 +170,9 @@ EDDSA (curve 25519), signature is 64 bytes.
 
 The consensus protocol version number in each block is monotonically increasing in the chain of blocks.
 
+A means - not under consensus - of activating upgrades to the consensus protocol
+is [miner signalling](./miner_signalled_consensus.md).
+
 #### Key Blocks
 
 ```
--- a/consensus/miner_signalled_consensus.md
+++ b/consensus/miner_signalled_consensus.md
@@ -32,7 +32,19 @@ The value in the field `info` in the key block headers in the
 - Uses the whole field `info` i.e. is not a bitmask.
 - Is meant to be a pseudorandom number, so to limit the risk of the same value
   being used by simultaneous competing consensus protocol proposals.
-- Is part of the validation rules of the new consensus protocol.
+- Is not necessarily part of the validation rules of the new consensus
+  protocol.
+  For the sake of clarity: the miner signalling is orthogonal to the
+  validation rules of the new consensus protocol.
+  This means that, if the network resonably settles - via miner
+  signalling - on activating the new consensus protocol and a user
+  starts a node configuring the new consensus protocol as
+  unconditionally effective at the same height as the (past)
+  miner-signalled consensus upgrade height, the node may - depending
+  on the chain forks observed on the network - follow as best valid
+  chain a chain fork that upgraded to the new consensus protocol at
+  the configured height even if without a majority of "signals" in the
+  "signalling" block interval.
 
 ### Parameters
 
@@ -101,6 +113,8 @@ of 3360) if (HE - HS) is 3360.
 
 ## Rationale
 
+### New consensus protocol activation per chain fork
+
 The node keeps the decision to activate the new consensus protocol local to the
 chain fork as taking that decision globally does not enhance the user experience
 and rather worsens it.
@@ -113,6 +127,14 @@ and rather worsens it.
 
 *NOTE: This design choice may be lifted if problematic during the implementation
  phase.*
+
+### Value of the info field not necessarily part of new consensus protocol
+
+The miner signalling is orthogonal to the validation rules of the new
+consensus protocol for the sake of simplicity of the implementations.
+
+This is not detrimental as the new consensus protocol may include
+arbitrary constraints.
 
 ## Backwards Compatibility
 
--- a/consensus/miner_signalled_consensus.md
+++ b/consensus/miner_signalled_consensus.md
@@ -24,8 +24,12 @@ the header field `info`.
 
 The "signalling" block interval has a determined number of confirmations before
 the determined height at which the new consensus protocol may become
-active. This is in order to prevent the node from having to aggregate the
-results of the "signalling" block interval in a too short time interval.
+active. This is in order to:
+- Prevent the node from having to aggregate the results of the
+  "signalling" block interval in a too short time interval.
+- Prevent a user from restarting the node configuring unconditionally
+  the new consensus protocol when the signalling outcome is not
+  confirmed with reasonably high probability.
 
 The value in the field `info` in the key block headers in the
 "signalling" block interval:
--- a/consensus/miner_signalled_consensus.md
+++ b/consensus/miner_signalled_consensus.md
@@ -1,48 +1,60 @@
 # Miner Signalled Consensus Upgrade
 
-A consensus protocol upgrade based on miner signalling outcome can be referred
-to as a _conditional_ consesus upgrade. A hard-fork (in the context of a
-consensus protocol upgrade based on miner signalling) happens with a key block,
-therefore, using the term "block" in the following text refers to a "key
-block". A consensus protocol upgrade not based on any condition is
-_unconditional_ protocol upgrade.
+## Abstract
 
-## Simple Summary
-
-This document describes how nodes in the network signal the readiness for a new
-consensus protocol, its activation, and a hard-fork at a certain height.
+This document describes miner signalled consensus upgrade. It is a mechanism
+that actives a new hard forking consensus on the æternity network when the
+majority of mining power supports it.
 
 ## Motivation
 
-Activating a new consensus protocol based on miner signalling outcome prevents
-the block chain to diverge in case:
-- A large fraction of the miners have not upgraded their nodes to a version
-able to follow the new consensus protocol.
-- The majority of the miners is against consensus protocol upgrade.
+There have been only *planned hard forks (unconditional)* that were well
+communicated in advance and include only consensus breaking features that the
+whole community agree with.
+
+Since the æternity is a complex platform that consists of components such as
+naming system, oracles, smart contracts, generalized accounts and state
+channels, the development of these components and the blockchain itself is
+likely to be consensus breaking in the future. Moreover, the Lima hard fork is
+the last planned hard fork, so planned hard forks must be replaced with another
+mechanism.
+
+To prevent the blockchain from a split, we chose to use the consensus upgrade
+based on the miner signalling outcome. The vast majority (90% or more) of mining
+power must be supporting the new consensus in order to upgrade the
+blockchain. The minority can still follow the old consensus, but since its
+mining power will be very low, this fork is generally supposed to become
+abandoned.
+
+Miner signalled consensus upgrade is a *conditional hard fork* since the
+activation of the new consensus is based on a condition. A hard-fork (in the
+context of a consensus protocol upgrade based on miner signalling) happens with
+a key block, therefore, using the term "block" in the following text refers to a
+"key block".
 
 ## Specification
 
 ### Overview
 
 On a node configured for a consensus upgrade based on miner signalling outcome,
-the new consensus protocol becomes active at a determined height if at least a
-determined majority of the blocks between two determined block heights
-(signalling interval) contains a predefined signal. Such consesus protocol
-upgrade happens per chain fork.
+the new consensus becomes active at a determined height if at least a determined
+majority of the blocks between two determined block heights (*signalling
+interval*) contains a predefined *signal*. Such consensus upgrade happens per
+each chain fork.
 
-The signal value is part of a block header in the signalling interval:
-- Is meant to be a unique number, so to limit the risk of the same value being
-  used by simultaneous competing consensus protocol proposals.
-- Is not necessarily part of the validation rules of the new consensus
+The signal value is part of a block header in the signalling interval. It:
+- is meant to be a unique number, so to limit the risk of the same value being
+  used by simultaneous competing consensus proposals;
+- is not necessarily part of the validation rules of the new consensus
   protocol. For the sake of clarity: miner signalling is orthogonal to the
-  validation rules of the new consensus protocol. This means that if the
-  network reasonably settles - via miner signalling - on activating the new
-  consensus protocol and a user starts a node configuring the new consensus
-  protocol as unconditionally effective at the same height as the (past) miner
-  signalled consensus protocol upgrade height, the node may - depending on the
-  chain forks observed on the network - follow as best valid chain a chain fork
-  that upgraded to the new consensus protocol at the configured height even if
-  without a majority of blocks with expected signal in the signalling interval.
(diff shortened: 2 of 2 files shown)


consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,159 @@
...
+  decision globally the node may anyway taint the storage with blocks with
+  eventually wrong consensus protocol.
+
+*NOTE: This design choice may be lifted if problematic during the implementation
+ phase.*
+
+## Backwards Compatibility
+
+The signalling using the block header field `info` is backward compatible, as
+the meaning of the value of such field is not under consensus.
```
sennui: `info` field value is under consensus, but the meaning of it - not. OK.

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,159 @@
...
+The height at which the new consensus protocol may be active. Let us call it H2.
+
+H2 should cater for the time required by the node for aggregating the results of
+the "signalling" block interval.
+
+H2 must be strictly greater than HE.
+
+The difference between H2 and HE is recommended to be approximately two
+hours i.e. 40 (`= 2 * 20`).
+
```
sennui: So we have: H1 < HS < HE < H2. OK.

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,159 @@
...
+#### Majority
+
+The number of key blocks in the "signalling" block interval that is meant to
+activate the new consensus protocol.
+
+It must be positive.
+
+It is recommended to be 90% of (HE - HS). So it is 3024 (that is - 90%
+of 3360) if (HE - HS) is 3360.
```
jur0: Is 90% going to be hardcoded or it's going be a configuration option?

lucafavatella: Configuration option.

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,159 @@
...
+The "signalling" block interval has a determined number of confirmations before
+the determined height at which the new consensus protocol may become
+active. This is in order to prevent the node from having to aggregate the
+results of the "signalling" block interval in a too short time interval.
+
+The value in the key block header field `info`:
+- Uses the whole field `info` i.e. is not a bitmask.
+- Is meant to be a pseudorandom number, so to limit the risk of the same value
+  being used by simultaneous competing consensus protocol proposals.
+- Is part of the validation rules of the new consensus protocol on the
```
lucafavatella: This may need to be reviewed/confirmed/refined.

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,159 @@
...
+The usage of the block header field `info` offers room for optimization for
+accommodating simultaneous hard-forking or soft-forking consensus proposals, by
+interpreting such `info` field as a bitmask rather than a value.
+
+### Soft-forking consensus proposals
+
+This proposal may be extended to soft-forking consensus proposals - as opposed
+to only hard-forking ones - by: relaxing the strict monotonicity of consensus
+versions; and excluding the value in the key block header field `info` in the
+"signalling" block interval from the validation rules of the new consensus
```
lucafavatella: ... and this may be revised too.

consensus/miner_signalled_consensus.md
```diff
@@ -186,5 +191,9 @@ then - with the same persisted storage - start (again) the node: the node
 needs not to consider any potential key block "signal" identified
 under the previous incorrect configuration.
 
+Any implementation of channels needs to ensure that the network
+reasonably settles on whether to activate the new consensus protocol
```
lucafavatella: Planned as https://www.pivotaltracker.com/story/show/165214367

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,196 @@
...
+per chain fork,
+the new consensus protocol becomes active at a determined
+height if at least a determined majority of the key blocks in a determined
+"signalling" block interval in such chain fork contains a determined value in
+the header field `info`.
+
+The "signalling" block interval has a determined number of confirmations before
+the determined height at which the new consensus protocol may become
+active. This is in order to prevent the node from having to aggregate the
+results of the "signalling" block interval in a too short time interval.
```
lucafavatella: And also for preventing a user seeing the new protocol active after the fork height to stop node and configure unconditionally the new consensus protocol, and then restarting node (that had other chain fork X) and that chain fork gets prolonged.

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,200 @@
...
+Any implementation needs to cater for the performance of the aggregation of the
+signalling e.g. slow multiple lookups on the storage.
+
+Any implementation needs to consider that the storage may contain blocks with
+the old and new consensus protocols at heights at which the new consensus
+protocol should have been applied. This is because the node should be aware of
+multiple chain forks and peers.
+
+Any implementation needs to be resilient to node restarts with
+accidental distinct miner signalling parameters in the configuration.
```
hanssv: Since any hard fork will mean a new "release"/version of the node the _only_ configuration is whether to signal or not. This does not affect the history in any way. I.e. there is nothing that needs to be re-computed when you restart the node. The (partial) result of the vote could then be tallied in every block (not under consensus, but in the local database) as the node goes along - this means there is no need for an additional worker process and there is no need to have a time window after the vote period ends. 

I think this will simplify things considerably, but maybe I am missing something...

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,200 @@
...
+
+Such extension should preferably be accompanied by a working implementation.
+
+## Implementation
+
+Any implementation needs to cater for the performance of the aggregation of the
+signalling e.g. slow multiple lookups on the storage.
+
+Any implementation needs to consider that the storage may contain blocks with
+the old and new consensus protocols at heights at which the new consensus
```
hanssv: The node has _one_ chain at any point in time, there will not be multiple key-blocks at the same height, I don't understand this?!

jur0:
```
                         [3] --- [4] --- [5] --- [6] --- [7] --- [8] ---  <- High difficulty
                          |
 [0] --- [1] --- [2] --- [3] --- [4] --- [5] --- [6] --- [7] --- [8] ---  <- Low difficulty
                  H1      HS            HE - 1    HE    H2 - 1    H2
```
There is just low difficulty chain at first and the fork signalling result is true (so switching to the new protocol at H2). Then there is a chain fork at height HS, this chain has higher difficulty, so it wins. The fork signalling result for high difficulty chain is false (so not switching to the new protocol at H2). Now there are 2 blocks in the db with different protocol versions at height H2 (but only the one in high difficulty chain is "active").

consensus/miner_signalled_consensus.md
jur0: The config should include more properties than "signal/not signal". There may be concurrent forks going on and we should make it possible to choose the signal value, etc.

If we wanted to keep the partial result in a local persistent db, then the result would apply just to one particular config. If we decide to restart the node, then the partial result might not be applicable anymore if we the config was changed. So we would need to recompute the result again and store it.

jur0: > there is no need for an additional worker process and there is no need to have a time window after the vote period ends.

The time window after the fork signalling interval ends is there mainly because we want the chain between HS and HE to reasonably settle (so there is less likelihood of forks in the signalling interval and so the fork signalling result at H2 shouldn't change)

consensus/consensus.md
```diff
@@ -170,6 +170,9 @@ EDDSA (curve 25519), signature is 64 bytes.
 
 The consensus protocol version number in each block is monotonically increasing in the chain of blocks.
 
+A means - not under consensus - of activating upgrades to the consensus protocol
+is [miner signalling](./miner_signalled_consensus.md).
```
ThomasArts: This sentence is hard to parse. Please clarify better

consensus/miner_signalled_consensus.md
hanssv: I don't think the signalling outcome should depend on the configuration... Either you decide to follow the outcome of the "poll" and then it is a feature that all nodes behave the same (since the poll is on chain they will) - if you don't want to follow the outcome you better run an older/other version of the node? **We** don't want the network to split, so let's not give them the tools for this ;-)

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,200 @@
+# Miner Signalled Consensus Upgrade
+
+## Simple Summary
+
+This document describes the proposal of how nodes in the peer-to-peer network
```
ThomasArts: --> This document describes a protocol to signal the readiness for, and activation of, a new hard-fork at a certain height.

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,200 @@
...
+
+Activating a hard-forking new consensus protocol based on miner signalling
+prevents diverging the block chain in case a large fraction of the miners have
+not upgraded their nodes to a version able to follow the new consensus protocol.
+
+## Specification
+
+### Overview
+
+On a node configured for a miner-signalled consensus upgrade,
```
ThomasArts: Nodes can be configured to handle a specific hard-fork as a reaction on signals from other nodes?

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,200 @@
...
+per chain fork,
+the new consensus protocol becomes active at a determined
+height if at least a determined majority of the key blocks in a determined
+"signalling" block interval in such chain fork contains a determined value in
```
ThomasArts: Explain the concept of "signalling block interval" first.

ThomasArts:
The text is written by mixing implementation details and general concepts and vision. It would be nicer to first explain the why and concepts and then give how and implementation details.

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,200 @@
...
+  unconditionally effective at the same height as the (past)
+  miner-signalled consensus upgrade height, the node may - depending
+  on the chain forks observed on the network - follow as best valid
+  chain a chain fork that upgraded to the new consensus protocol at
+  the configured height even if without a majority of "signals" in the
+  "signalling" block interval.
+
+### Parameters
+
+In order for the miner-signalled consensus protocol upgrade to succeed, each
```
ThomasArts: If each node needs to have the same set, the this set better is under consensus

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,200 @@
...
+The proposed new consensus protocol version.
+
+It must be strictly greater than C1 (strictly monotonically increasing consensus
+protocol version).
+
+#### Signal
+
+The "signal" value in the block header field `info`.
+
+It should be pseudorandom.
```
ThomasArts: what does that mean?

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,200 @@
...
+
+#### Signalling block interval start
```
ThomasArts: better to start with this earlier in the story

consensus/miner_signalled_consensus.md
jur0: The configuration values would have to be hardcoded in the node then, and the only configuration option would be "follow signalling outcome/stay with the current protocol version".

I still prefer having the fork signalling configuration in the configuration file though. This is more flexible and it would allow the community to do protocol upgrade themselves (which may not be the best thing from our point of view). My understanding of fork signalling is that we just provide means of doing the protocol upgrades, we suggest some features and let the community decide. Maybe some other party can suggest new features as well, but they still will be able to use the fork signalling mechanism since it will be configurable.

This might be something we need to discuss further and more people should be involved...

consensus/miner_signalled_consensus.md
jur0: It's to make sure that in case of concurrent forks - meaning that aeternity will release a node which will try to upgrade to a new protocol, but also another party will do the same so the two signals will not be the same.

consensus/miner_signalled_consensus.md
jur0: It means that the fork signalling needs to be fork aware. It's showed in https://github.com/aeternity/protocol/pull/385#discussion_r325211457

So if the protocol upgrade happens on the current chain, it doesn't necessarily happen on all chain forks.

consensus/miner_signalled_consensus.md
jur0: Please have a look a the comments [here](https://github.com/aeternity/protocol/pull/385/files#r325690165).

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,247 @@
+# Miner Signalled Consensus Upgrade
```
knarz: In general I feel like this document is not so much about the miner signalling and more about how a node deals with the signals. It might help to clear that up?

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,247 @@
...
+height). This is in order to:
+- Prevent the node from having to aggregate the results of the signalling
+  interval in a too short time interval.
+- Prevent a user from restarting the node and configuring the new consensus
+  protocol unconditionally when the signalling outcome is not confirmed with
+  reasonably high probability.
+
+The signal value is part of a block header in the signalling interval:
+- Is meant to be a pseudorandom number, so to limit the risk of the same value
+  being used by simultaneous competing consensus protocol proposals.
```
knarz: What if someone chooses the same value intentionally?

consensus/miner_signalled_consensus.md
```diff
@@ -0,0 +1,247 @@
...
+The height at which the signalling interval starts - included. Let us call it
+HS.
+
+HS should cater for the time required by most willing miners to adopt a version
+of the node able to signal and able to follow the new consensus protocol
+e.g. at least two weeks.
+
+HS must be strictly greater than H1.
+
+#### Signalling interval end
```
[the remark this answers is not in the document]
jur0: The implementation doesn't allow anything to be configurable for mainnet and testnet. The only option is "follow miner signalling outcome" or "stay with the current protocol". Other environments allow configuration of all the signalling parameters.

consensus/miner_signalled_consensus.md
jur0: When the majority of the mining power signals a new protocol and then also follows it, then the nodes that were signalling the new protocol, but now are not able to follow it, will end up on a forked chain.

dincho:
I'm afraid this is outdated.

PR #386, opened 2019-07-30
update ECDSA example with latest Sophia syntax

mradkov:
update ECDSA example with latest Sophia syntax

- removed custom `require` function as it is now available in Sophia
- updated function naming with the latest `entrypoint` / `function` syntax for public and private functions


pacaloka:
Thanks

hanssv:
Fixing merge conflict in another PR....

PR #392, opened 2019-08-05
added subname transaction

skkw:


AENS.md
```diff
@@ -90,7 +90,7 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 ### AENS Entry
 
-This is the entry as it should be stored by a node.
+This is the **top** name entry as it should be stored by a node.
```
lucafavatella: I read above:

> Any label that is not at the top level, e.g. `aet`, ...

So a "top" name is e.g. "aet".

So this paragraph refers to names without any dots in them?

AENS.md
```diff
@@ -125,7 +125,19 @@ This can have multiple entries, e.g. a payment address associated
 with the name and an oracle address associated with the name.
 
 
-### Name
+Entry for **subname** (or, subdomain) uses all the fields from
```
lucafavatella: What is a "subname"?

Is a subname an entity already included - via the other transactions - in the Fortuna release of AENS? (Clarifying this may require you to open a separate preparatory PR against master in which you detail Fortuna AENS.)

AENS.md
skkw: My understanding is that "aet" or "test" are called "registrars" in the document. 

"Any label..." - I'm not entirely sure whether I should correct things already in document or focus on subnames description. I think it makes sense to restrict ourselves to subnames description here, and possibly make another correction PR which could focus on fixing ambiguities in the whole text.

AENS.md
```diff
@@ -347,6 +376,30 @@ the name enters the `revoked` state. After a fixed timeout of
...
+
+The `subname` transaction MUST be signed by the owner of the `name` entry.
+
+`subname` transaction is used to define a tree of subnames below `name` entry
+along with the pointers for each subname.
+
+A `definition` field is a mapping of `subname` to `pointers`, where the `pointers`
+for subname SHOULD NOT contain multiple entries with the same key.
+
+The execution of `subname` transaction first removes all subnames belonging to provided
```
lucafavatella: May be made clearer (as only observable outcome is root hash of tree - included in block header). Though minor.
```
The execution of `subname` transaction is semantically equivalent to ....
```

serializations.md
```diff
@@ -527,6 +529,17 @@ The content of the contract store depends on [the ABI and the VM version](/contr
...
+[ <account_id> :: id()
+, <nonce>      :: int()
+, <name>       :: binary()
+, <definition> :: [{binary(), [{binary(), binary()}]}]
+, <fee>        :: int()
+, <ttl>        :: int()
+]
+```
+
 The recipient must be one of the following:
```
lucafavatella: Misplaced section? 🤔

AENS.md
```diff
@@ -125,7 +125,36 @@ This can have multiple entries, e.g. a payment address associated
 with the name and an oracle address associated with the name.
 
 
-### Name
+A subname entry (or, subdomain), belongs to a name entry.
```
lucafavatella: This document shall be clarified on what is for Lima release onwards.

AENS.md
```diff
@@ -347,6 +376,30 @@ the name enters the `revoked` state. After a fixed timeout of
...
+
+A `definition` field is a mapping of `subname` to `pointers`, where the `pointers`
+for subname SHOULD NOT contain multiple entries with the same key.
+
+The execution of `subname` transaction first removes all subnames belonging to provided
+existing name, and then inserts all subnames into naming tree with their respective pointers.
+
+To just remove all subnames of a name, the `definition` should be empty.
+
+
```
lucafavatella:
```suggestion
NOTE: The size and complexity of the configuration of subnames is implicitly capped by the block gas limit.

```

AENS.md
```diff
@@ -347,6 +376,30 @@ the name enters the `revoked` state. After a fixed timeout of
...
+ ------------ ----
+| name       | 63 |
+ ------------ ----
+| definition |    |
+ ------------ ----
+```
+
+The `subname` transaction MUST be signed by the owner of the `name` entry.
+
+`subname` transaction is used to define a tree of subnames below `name` entry
```
lucafavatella: If I read this correctly, "tree of subnames" means no subname without a parent. Ok.

AENS.md
```diff
@@ -347,6 +376,30 @@ the name enters the `revoked` state. After a fixed timeout of
...
+along with the pointers for each subname.
+
+A `definition` field is a mapping of `subname` to `pointers`, where the `pointers`
```
lucafavatella: If no subname is meant to be without parent, then this needs to be detailed for preventing attacks. I am focusing on the semantics layer - not on the deserialization layer.

A. If constraint on parenthood enforced strictly, then malicious user may post subname tx filled with subnames up to microblock gas limit, and having as last name an orphan subname. What is behaviour in this case? If dropping all tx (hence not charging sender), this does not account for the computational cost of the past large portion of the list of subnames. Recommended solution: add a SHOULD constraint on subnames being a tree, noting that subnames that are orphan are silently ignored (the idea is so that user gets charged for computational cost).

B. If tx needs to be processed fully in order to know whether subname shall be added or not, this opens the door for tx computational cost non-linear with size of tx. Recommended solution: add a SHOULD constraint on subnames being sorted according to a specific algorithm you specify here (TO-BE-DONE) for making computational cost of check more predictable.

C. Can subnames have arbitrary depth? Please comment on max depth (explicit or implicit - if any) and how that impacts linearity of gas cost of processing this transaction at the increase of the depth of the names.

serializations.md
skkw: fixed

AENS.md
skkw: fixed in latest commits

AENS.md
skkw: added

AENS.md
skkw: It's now handled in implementation - we don't create intermediate nodes in aens tree. Processing cost of the list of subnames with pointers should be linear to the tx size.

skkw:
> Unclear definitions/entities, lifecycle, semantics of existing and new transactions.
> 
> Unclear [gas costs](https://github.com/aeternity/protocol/blob/4e1bbe586c68665c9a24ecb40582f99af3bec894/consensus/consensus.md#gas).

Added in latest commit.

Sophia language integration is postponed for later.

skkw:
> Brilliant! I requested a couple of clarifications.
> 
> AEVM primops?
> FATE VM hooking?
> Sophia language?

Postponed a bit to decouple TX and its integration.

AENS.md
skkw: Clarified in latest commit.

lucafavatella:
> > Brilliant! I requested a couple of clarifications.
> > AEVM primops?
> > FATE VM hooking?
> > Sophia language?
> 
> Postponed a bit to decouple TX and its integration.

@skkw  Could you please post the Pivotal Tracker ticket tracking this? Thanks.

AENS.md
lucafavatella: See also https://github.com/aeternity/protocol/pull/392#discussion_r311969444

serializations.md
```diff
@@ -529,6 +529,12 @@ The content of the contract store depends on [the ABI and the VM version](/contr
 ]
 ```
 
+The recipient must be one of the following:
```
lucafavatella: I understand correctly this is clarification that applies to Fortuna already. Up to you if you would prefer to split it as separate PR against master.

AENS.md
```diff
@@ -347,6 +381,32 @@ the name enters the `revoked` state. After a fixed timeout of
...
+for subname SHOULD NOT contain multiple entries with the same key.
+
+The execution of `subname` transaction is semantically equivalent to:
+- removal of all subnames belonging to provided `existing` name
+- inserting of all subnames into naming tree with their respective pointers
```
lucafavatella: Unclear what behaviour is for duplicated subname.

AENS.md
```diff
@@ -347,6 +381,32 @@ the name enters the `revoked` state. After a fixed timeout of
...
+ ------------ ----
+| name       | 63 |
+ ------------ ----
+| definition |    |
+ ------------ ----
+```
+
+The `subname` transaction MUST be signed by the owner of the `name` entry.
+
+`subname` transaction is used to define a tree of subnames below `name` entry
```
lucafavatella: Ambiguous. A "tree" may be understood as implying that each subname has a parent from label point of view.

AENS.md
```diff
@@ -125,7 +125,36 @@ This can have multiple entries, e.g. a payment address associated
...
+they belong to.
+
+Issuing `revoke` or `transfer` transactions on name causes revoking or
+transfering of all of the name's subnames.
+
+When the name enters `unclaimed` state, all of the name's subnames
+are deleted.
+
+Issuing `claim`, `revoke`, `transfer` and `update` transaction where
+subname is provided instead of name causes failure, and such transaction
```
ThomasArts: Propagation to other nodes is out of the context of this document. This is just about valid and invalid transctions.
So, better:
"such transactions do not validate".

AENS.md
```diff
@@ -125,7 +125,36 @@ This can have multiple entries, e.g. a payment address associated
...
+is not propagated to other nodes.
+
+The only subname specific fields are **pointers** which can be
+unique for every subname entry.
```
ThomasArts: I rather see a better specification here. Somehow they MUST be unique, because if you give mpultiple entries the implementation choses one of them. Also describing what a point is could be good.

AENS.md
```diff
@@ -148,6 +177,11 @@ interoperability with DNS.
 
 ***Note:*** For this first draft names MUST `.test` as a suffix, i.e. `mywallet.test`.
 
+Subnames are subject to the same restriction applied to names, with one
```
ThomasArts: It is discussible if we would not see google.test as a subname of the test domain. Clearly not, since it is a name, but the line is thin here

dincho:
I'm afraid this is outdated.

PR #394, opened 2019-08-06
Add backchannel updates to channels_api_usage.md

uwiger:
Documentation PR for https://github.com/aeternity/aeternity/pull/2611

Includes updated examples from the improved client reconnect test cases.


AleksandarFilipov:
I'm late in the game here but would like to suggest that the message received by the connected client isn't `channels.conflict`. Instead I would have preferred some more detailed message like - `channels.client.absent`.

uwiger:
> I'm late in the game here but would like to suggest that the message received by the connected client isn't `channels.conflict`. Instead I would have preferred some more detailed message like - `channels.client.absent`.

Yes, this should be captured in an issue.

PR #396, opened 2019-08-08
[PT-164136571] State channel cache encryption

grzegorz225:
Documentation for https://github.com/aeternity/aeternity/pull/2630.

### DIFF
--- a/channels/OFF-CHAIN.md
+++ b/channels/OFF-CHAIN.md
@@ -400,7 +400,12 @@ In the Aeternity node implementation, the state trees are cached inside the node
 Note that if the node restarts, cached data is not likely to survive (it is
 not persisted on each update for performance reasons.) The Aeternity node channel
 fsm automatically recovers the state trees and verifies that the provided
-state is in fact the last mutually authenticated state.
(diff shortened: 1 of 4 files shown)


channels/OFF-CHAIN.md
```diff
@@ -400,7 +400,12 @@ In the Aeternity node implementation, the state trees are cached inside the node
 Note that if the node restarts, cached data is not likely to survive (it is
 not persisted on each update for performance reasons.) The Aeternity node channel
 fsm automatically recovers the state trees and verifies that the provided
-state is in fact the last mutually authenticated state.
+state is in fact the last mutually authenticated state. In case the FSM dies 
+or an participant leaves the channel the state trees are persisted on disk and protected with
+an user provided passwords. For security reasons the password must be at least 6 unicode characters
+long. After the lima hard fork providing a state password will become mandatory - before the lima fork
+a default password is used - `correct horse battery staple`.
```
lucafavatella: Is this not user API breaking so unable to be merged before 5.0.0?

This also mean that doc and code can be simplified, making password mandatory, and maybe removing some error codes?

PR #399, opened 2019-08-13
Mentioned standard library and list comprehensions

radrow:
This waits for https://github.com/aeternity/aesophia/pull/105 to be merged.

While you are inspecting this PR – where should docs for standard library go? Imo here they will take too much space in sophia.md...


hanssv:
Let's make a separate file for the standard lib... Also we should re-think the implicit inclusion of them, but that is a separate concern...

PR #401, opened 2019-08-14
Fixed Sophia Syntax Lexical definition

robert-zaremba:
in the `Decl` definition the stateful should be a final term.


hanssv:
This will be fixed by #403 - and they would conflict, so closing this one.

Thanks for the contribution!

PR #409, opened 2019-08-30
Update sophia.md safety information

robert-zaremba:
We often like to search documentation through keywords, like `overflow`. 
I've extended arithmetic operations documentation by adding information about `integer overflow`.


UlfNorell:
Superseded by #413.

robert-zaremba:
Hi @UlfNorell .
Good suggestion.
PS: I would be happy to have my commit as a part of the contribution effort, instead of being superseded.

PR #416, opened 2019-09-18
Native tokens RFC WIP

happi:
First draft of Native tokens

### DIFF
--- a/serializations.md
+++ b/serializations.md
@@ -141,6 +141,7 @@ ambiguity.
 | 4      | oracle          |
 | 5      | contract        |
 | 6      | channel         |
+| 7      | token           |
 
 In Erlang notation, the `id()` type pattern is:
 ```
@@ -252,6 +253,7 @@ subsequent sections divided by object.
 | Sophia byte code | 70 |
 | Generalized accounts attach transaction | 80 |
 | Generalized accounts meta transaction | 81 |
+| Tokens | 90 |
 | Key block | 100 |
 | Micro block | 101 |
 | Light micro block | 102 |
@@ -280,6 +282,14 @@ subsequent sections divided by object.
 , <nonce>       :: int()
 , <balance>     :: int()
 ]
+```
+
+#### Accounts (version 4, Normal accounts with flags and tokens, from XXX release)
+```
+[ <flags>       :: int()
+, <nonce>       :: int()
+, <balance>     :: int()
+, <tokens>      :: [{id(),int()}]
 ```
 
 ### Signed transaction
--- a/serializations.md
+++ b/serializations.md
@@ -319,6 +319,28 @@ The recipient must be one of the following:
 * A contract identifier.
 * A name identifier, whose related name entry has an identifier as value of pointer with key `account_pubkey`.
   If multiple pointer entries are present for such key, then the first of such entries is used.
+`
+
+### Spend token transaction
+```
+[ <sender>    :: id()
+, <recipient> :: id()
+, <token>     :: id()
+, <amount>    :: int()
+, <fee>       :: int()
+, <ttl>       :: int()
+, <nonce>     :: int()
+, <payload>   :: binary()
+]
+```
+
+The recipient must be one of the following:
+* An account identifier.
+* An oracle identifier.
+* A contract identifier.
+* A name identifier, whose related name entry has an identifier as value of pointer with key `account_pubkey`.
+  If multiple pointer entries are present for such key, then the first of such entries is used.
+`
 
 #### Oracles
 ```
--- a/tokens/tokens.md
+++ b/tokens/tokens.md
@@ -1,0 +1,67 @@
+# Native Tokens
+
+A native tokens is made up of two parts: a Native token type and a native token balance.
+
+Native tokens types are a first class object on the chain that defines the
+behaviour of a type of tokens. Native tokens live in their own account balances
+of accounts.
+
+Native tokens (from now on just tokens) are:
+- Transferable (`spend`, `trade`)
+- Countable
+- Mintable (`mint`)
+- Destroyable (`burn`)
+
+Token types are:
+- Creatable
+
+
+## Token transactions
+
+There is one transaction on token types:
+- create
+
+There are five token transactions:
+- mint
+- spend
+- trade
+- burn
+- finalize
+
+Other transactions that takes a balance can also take a native token balance.
+
+### token create transaction
+
+The token create transaction takes the following argument
+- Meta data : string
+and the followin optional arguments
+- Contract
+- Amount
+- Recipient
+- Parent
+- Final
+
+The `meta data` is an uniterpreted string but token minters are encouraged to
+use the following json object:
+{
+  "type" : "object",
+  "properties": {
+    "name": {"type" : "string"},
+  }
+}
+
+The `contract` has to provide the following ACI:
+ spend(recipient : address, payload : Type) : booleano
+ trade([(from : address, to: address, Optional(token : address))], payload : Type)
+ mint(amount: integer)
+ burn(amount : integer) : boolean
+
+The `amount` is the number of tokens to create.
+
+The `recipient` an account to dump the tokens created, if not given the tokens are
+spent to the creators account.
+
+The `parent` is a pointer to a parent token for hierarchical tokens.
+
+The `final` argument sets the final flag which would block future minting.
+
--- a/tokens/tokens.md
+++ b/tokens/tokens.md
@@ -51,7 +51,7 @@ use the following json object:
 }
 
 The `contract` has to provide the following ACI:
- spend(recipient : address, payload : Type) : booleano
+ spend(recipient : address, payload : Type) : boolean
  trade([(from : address, to: address, Optional(token : address))], payload : Type)
  mint(amount: integer)
  burn(amount : integer) : boolean
@@ -65,3 +65,8 @@ The `parent` is a pointer to a parent token for hierarchical tokens.
 
 The `final` argument sets the final flag which would block future minting.
 
+If a contract is provided then any transaction (spend, trade, mint, burn) would
+call this contract and the transaction only goes through if the result of the
+corresponfing contract call returns true. Any other transaction using
+the token (such as contract call) would only be ececuted if a call to spend
+returns true.


tokens/tokens.md
```diff
@@ -0,0 +1,99 @@
...
+A native tokens is made up of two parts: a Native token type and a native token balance.
+
+Native tokens types are a first class object on the chain that defines the
+behaviour of a type of tokens. Native tokens live in their own account balances
+of accounts.
+
+Native tokens (from now on just tokens) are:
+- Transferable (`spend`, `trade`)
+- Countable
+- Mintable (`mint`)
```
sennui: Who can issue mint transaction? I imagine, token issuers would like to customize minting for their needs.

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+
+Informally, when you need a new type of token for whatever purpose,
+you create an ANT describing the token, and then you can mint new
+tokens and start sending them to accounts.
+
+Owners of the tokens can spend them to other accounts, or trade them
+(atomically) for other assets with other accounts. If tokens are no
+longer wanted you can burn them.
+
+Only the owner of the ANT can mint new tokens, and is the owner
```
ThomasArts: --> "Only the owner of the ANT can mint new tokens. Only the owner
decides the total amount of token. The ANT can be finalized, preventing " ...

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+decides that enough is enough, the ANT can be finalized, preventing
+more minting. An ANT can also be created with a final supply already
+minted and given to the owner of the ANT.
+
+An ANT can only be destroyed (TODO: design decision) if the total
+token supply of the ANT is zero (i.e., all tokens are burnt).
+
+Fees cannot be payed by native tokens, but only by aeons, but other
```
ThomasArts: --> "Fees cannot be paid by native tokens, but only by aeons. However, other
services can be paid for with native tokens. (TODO: Oracle query fees, etc?)" ...

ThomasArts: Somewhere it has to be made clear that "aeons" are not tokens. If one were able to trade tokens to aeons and back, one is most likely able to create new aeons in the system, which we don't want.

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+other transaction that can pass an amount. You can:
+- spend tokens through the `spend_tx`
+- pass tokens as value in a `contract_call_tx` or `contract_create_tx`
+- pass tokens as value in contract calls in a smart contract.
+- spend tokens as query fees in `oracle_query_tx` (TODO: Might be a future extension).
+
+### ANT create transaction (`ant_create_tx`)
+
+The ANT create transaction takes the mandatory argument:
+- Meta data : string
```
ThomasArts: This is too weak. I think each token must have a unique identifier, say a 32 byte hash or so. Just a string is hard to handle in contracts.

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+- Parent  (TODO: Decide what impact this have. Possibly delay having hierarchical tokens)
+- Final
+
+The `meta data` is an uninterpreted string but token minters are encouraged to
+use the following json object:
+```
+{
+  "type" : "object",
+  "properties": { "name": {"type" : "string"},
+                }
```
ThomasArts: and an identifier... sure, that could be the hash of the string, but it has to be the unique thing to work with in the contract. We could even allow a string of 3 unique characters, like in valutas... that's limited, but makes contracts very readable.

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+}
+```
+
+The `contract` has to provide the following ACI:
+```
+ spend(recipient : address, payload : Type) : boolean
```
ThomasArts: What is the payload type here? A binary string, I guess, like in normal spend?

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+ trade([(from : address, to: address, Optional(token : address))], payload : Type) : boolean
```
ThomasArts: Ok, the token has an address... that's the uqniue identifier. "tk_...." or something in the contract.

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+ mint(amount: integer) : boolean
+ burn(amount : integer) : boolean
+```
+
+The `amount` is the number of tokens to create.
```
ThomasArts: Some link has to be established between contract attached to tokens and ANT itself. If one starts an ANT with a specific amount and not allow to create more tokens, then the contract should respect this. I would say that the ANT specification always has preference over contract specification.

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+
+The `recipient` an account to dump the tokens created, if not given the tokens are
```
ThomasArts: ???

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+spent to the creators account.
+
+The `parent` is a pointer to a parent token for hierarchical tokens.
+
+The `final` argument sets the final flag which would block future minting.
+
+If a contract is provided then any transaction (spend, trade, mint, burn) would
+call this contract and the transaction only goes through if the result of the
+corresponfing contract call returns true. Any other transaction using
```
ThomasArts: --> "corresponding"

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+the token (such as contract call) would only be executed if a call to spend
+returns true.
+
+## Tokens state tree
+
+The tokens state tree contain token type objects.
```
ThomasArts: where the type object is a bit underspecified at the moment... but that's fine.

tokens/tokens.md
```diff
@@ -0,0 +1,144 @@
...
+
+- The token state tree contains
+  - Token type objects
+
+#### The token type object
+
+- Created by an token create transaction.
+
+```
+{ creator         :: id()
```
ThomasArts: So what is the actual token id?

tokens/tokens.md
gorillainduction: The identifier of the token should be a hash of something. Perhaps the creator's account and nonce, just as in oracle queries.

The meta data is not intended to be the identifier. This should be specified here, of course.

tokens/tokens.md
gorillainduction: I think it probably could be something ABI encoded. Just as in oracles. :thinking:

tokens/tokens.md
gorillainduction: Indeed.

tokens/tokens.md
gorillainduction: Agreed. Didn't come there in the rewrite yet ;-)

tokens/tokens.md
gorillainduction: Well, aeons are tokens, but not controlled by an ANT. You should surely be able to trade aeons for native tokens. There will not be new aeons created. Two parties simply agrees to make the trade, so the aeons come from one of the parties.

tokens/tokens.md
```diff
@@ -0,0 +1,203 @@
...
+```
+  id := Blake2b(<CreatorPubkey><CreateTxNonce>)
+```
+where `Blake2b` is the 256 bit hash. The [API
+serialization](../node/api/api_encoding.md) of an ANT id is tagged by
+`nt_`
+
+The ANT contains the fields:
+- `creator` - The account id of the creator
+- `meta_data` - A byte array field uninterpreted under consensus
```
hanssv:
```suggestion
- `meta_data` - A byte array field, uninterpreted, but under consensus
```

serializations.md
```diff
@@ -282,6 +289,14 @@ subsequent sections divided by object.
 ]
 ```
 
+#### Accounts (version 4, Normal accounts with flags and tokens, from XXX release)
+```
+[ <flags>       :: int()
```
hanssv: So a Generalized account can't own tokens?

tokens/tokens.md
```diff
@@ -0,0 +1,203 @@
...
+- Meta data : string
+- Amount
+- Recipient
+- Final
+- Contract
+- Parent  (TODO: Hierarchical tokens?)
+
+The `amount` is the number of tokens to mint at create time. Set to
+`0` if none should be minted. This can for example be combined with
+setting the `final` argument to false, thereby immediately minting all
```
hanssv: do you mean `true`?

serializations.md
gorillainduction: Will fix

tokens/tokens.md
gorillainduction: Yes

tokens/tokens.md
```diff
@@ -0,0 +1,207 @@
...
+- amount
+- recipient
+- final
+
+Only the `owner` can mint new tokens, but it can pass the minted
+tokens to a recipient. If `final` is set to true, the ANT will be
+finalized after the new tokens are minted.
+
+### ANT destroy transaction
+TODO: Should we be able to destroy an ANT that has a `total_supply` of 0?
```
UlfNorell: What would be the incentive for the owner to destroy the ANT? Also it seems unlikely that the total supply will ever reach 0 (case-in-point: aeternity token migration).

tokens/tokens.md
```diff
@@ -0,0 +1,207 @@
...
+If a contract is provided, any transaction (spend, trade, mint, burn)
+would call this contract and the transaction only goes through if the
+result of the corresponding contract call returns true. Any other
+transaction using the token (such as contract call) would only be
+executed if a call to spend returns true. (TODO: Decide how this plays
+with contract calls that tries to pass tokens as value, etc).
+
+Note that the `final` field in the ANT takes precedence over the
+governing contract. If the ANT is final, no minting can occur.  (TODO:
+Perhaps the contract should have a `finalize` endpoint as well?)
```
UlfNorell: 👍

tokens/tokens.md
```diff
@@ -0,0 +1,207 @@
...
+The contract may contain other endpoints as well, but at least one of
+the entrypoints above must be implemented. The contract can only be
+attached at create time.
+
+If a contract is provided, any transaction (spend, trade, mint, burn)
+would call this contract and the transaction only goes through if the
+result of the corresponding contract call returns true. Any other
+transaction using the token (such as contract call) would only be
+executed if a call to spend returns true. (TODO: Decide how this plays
+with contract calls that tries to pass tokens as value, etc).
```
UlfNorell: Who pays for the gas (the caller of the transaction presumably)? This means that all transactions dealing with tokens need gas/gas price fields.

tokens/tokens.md
```diff
@@ -0,0 +1,207 @@
...
+                }
+}
+```
+
+A contract governing the usage of tokens must have a non-empty subset
+of the following ACI:
+
+```
+ spend(recipient : address, payload : Type) : boolean
+ trade([(from : address, to: address, Option(token : address))], payload : Type) : boolean
```
UlfNorell: amounts?

tokens/tokens.md
```diff
@@ -0,0 +1,207 @@
...
+
+### ANT create transaction (`ant_create_tx`)
+
+The ANT create transaction takes the argument:
+- Meta data : string
+- Amount
+- Recipient
+- Final
+- Contract
+- Parent  (TODO: Hierarchical tokens?)
```
UlfNorell: What does hierarchical tokens mean?

tokens/tokens.md
```diff
@@ -0,0 +1,207 @@
...
+`account`. The burned amount is also counted from the `total_supply`
+in the ANT object.
+
+### Other transactions on tokens
+
+Tokens can also be transfered from one account to another through any
+other transaction that can pass an amount. You can:
+- spend tokens through the `spend_tx`
+- pass tokens as value in a `contract_call_tx` or `contract_create_tx`
+- pass tokens as value in contract calls in a smart contract.
```
UlfNorell: Consider allowing multiple currencies (for instance, some tokens + a tip in aeons).

tokens/tokens.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+A contract governing the usage of tokens must have a non-empty subset
+of the following ACI:
+
+```
+ spend(recipient : address, payload : Type) : boolean
+ trade([(from : address, to: address, Option(token : address))], payload : Type) : boolean
+ mint(amount: integer) : boolean
+ burn(amount : integer) : boolean
```
UlfNorell: `Call.caller` = the account burning the tokens, presumably. And same for `spend`.

marc0olo:
cool to see things happening around the native token support! :+1:

tokens/tokens.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+The contract may contain other endpoints as well, but at least one of
+the entrypoints above must be implemented. The contract can only be
+attached at create time.
+
+If a contract is provided, any transaction (spend, trade, mint, burn)
+would call this contract and the transaction only goes through if the
+result of the corresponding contract call returns true. Any other
```
UlfNorell: Note that in the case of a trade there will be one contract call per unique ANT used in the trade. What if different ANTs share the same contract?

tokens/tokens.md
```diff
@@ -0,0 +1,207 @@
...
+denominator, a display name, etc).
+```
+{
+  "type" : "object",
+  "properties": { "name": {"type" : "string"},
+                }
+}
+```
+
+A contract governing the usage of tokens must have a non-empty subset
```
thepiwo: but there must not be a governing contract?

PR #430, opened 2019-10-30
Update sophia.md

alexpanter:
I found the description of a list a bit ambiguous, so I have added some notes to start a small discussion. Feel free to delete/ignore if these are not relevant to you, or in case I misunderstand your design thoughts.


radrow:
Shouldn't this be an issue?

radrow:
This file is no longer in this repo (was moved to aesophia/docs).

PR #431, opened 2019-10-31
Update hashing name specification for Lima

sennui:


AENS.md
```diff
@@ -249,7 +252,10 @@ def namehash(name):
     return Blake2b(namehash(remainder) + Blake2b(label))
 ```
 
-Names are generally only referred to in hashed form.
+#### From Lima hardfork
+
+From Lima we don't use `NameHash` anymore. Instead, it is a 32 byte Blake2b hash
+of the UTF8 name that is IDNA encoded.
```
hanssv:
```suggestion
of the IDNA encoded UTF-8 name.
```

PR #432, opened 2019-11-03
update int type for fate

thepiwo:


thepiwo:
Yes, I agree there should be more documentation on the FATE/AEVM differences here

radrow:
Should be ok right now?

PR #433, opened 2019-11-04
Clarify potential values in aens pointers

sennui:


AENS.md
```diff
@@ -464,6 +464,8 @@ the future.
 We do not require an additional fee for extending the lease.
 
 The `pointers` field SHOULD NOT contain multiple entries with the same key.
+The `pointers` can point to one of: account address, oracle id,
```
tolbrino: The spec above should clarify that a list of pointer k/v entries is expected. Right now its just empty which is confusing.

sennui: @tolbrino true. The problem is partially that the format of this documentation mixes meaning behind protocol and syntax level documentation. `k/v` description has to be followed up with list of `k`s currently supported. But that's correct, I will add it in follow up PR.

ThomasArts: I think it is important to keep the SHOULD NOT, since if one points to the same key, only one of them is taken.

PR #436, opened 2019-11-05
Describe transaction TTL

hanssv:


noandrea:
Does this means that the TTL is related to a transaction already included in a block or is it related to a transaction waiting in the mempool?

hanssv:
> Does this means that the TTL is related to a transaction already included in a block or is it related to a transaction waiting in the mempool?

Once the transaction is included in a block the TTL has no meaning. This is a way to post a transaction to the mempool and make sure that unless it is included before block X it won't be included at all.

PR #443, opened 2019-11-13
GH-2851 Docs for PayingForTx

hanssv:
Associated to https://github.com/aeternity/aeternity/pull/2999

Also fixes some obviously incorrect values.

### DIFF
--- a/consensus/consensus.md
+++ b/consensus/consensus.md
@@ -282,6 +282,23 @@ both the `Fee` and the `GasPrice`.
  -------------- -----
 ```
 
+#### PayingFor transaction
+Available from protocol version 5, Iris release. The `tx` field size
+depends on the size of the inner transaction.
+```
+ Fieldname       Size (bytes)
+ -------------- -----
+| payer        | 32  |
+ -------------- -----
+| fee          | var |
+ -------------- -----
+| nonce        | 8   |
+ -------------- -----
+| tx           | var |
+ -------------- -----
+```
+
+
 #### Contract transactions
 
 Please refer to the [dedicated document](/contracts/contract_transactions.md).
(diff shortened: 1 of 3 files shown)


consensus/consensus.md
```diff
@@ -274,14 +274,31 @@ both the `Fee` and the `GasPrice`.
  -------------- -----
 | recipient    | 32  |
  -------------- -----
-| amount       | 8   |
+| amount       | var |
```
tolbrino: `var` should be defined somewhere.

consensus/consensus.md
```diff
@@ -339,7 +356,9 @@ The gas of a transaction is the sum of:
...
 | Contract call (AEVM)   | `30 * BaseGas` | Proportional to the byte size of the transaction, specifically: `byte_size(ContractCallTx) * GasPerByte`. It also includes gas for contract execution. |
 | GA Attach              | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAAttachTx) * GasPerByte`. It also includes gas for execution of the init function. |
-| GA Meta                | `5 * BaseGas`  | Proportional to the byte size of the transaction, specifically: `byte_size(GAMetaTx) * GasPerByte`. It also includes gas for execution of the authentication function + recursively gas corresponding to the wrapped transaction(s) (excluding the byte size portion - in order not to account for the size of wrapped transactions multiple times). |
+| GA Meta                | `5 * BaseGas`  | Fortuna and Lima : Proportional to the byte size of the transaction, specifically: `byte_size(GAMetaTx) * GasPerByte`. It also includes gas for execution of the authentication function + recursively gas corresponding to the wrapped transaction(s) (excluding the byte size portion - in order not to account for the size of wrapped transactions multiple times). |
```
tolbrino: Reads like this is not available post-Lima.

serializations.md
```diff
@@ -995,6 +996,20 @@ authorization data.
 ]
 ```
 
+### PayingFor
+The `PayingFor` transaction is available from version 5, Iris release. By using
+it an account `P` can pay for the transaction (transaction fee + gas) of another
```
ThomasArts:
```suggestion
it,  an account `P` can pay for the transaction (transaction fee + gas) of another
```

consensus/consensus.md
hanssv: Hmm, I just saw it being used further up in the same document so I assumed it to be defined...

consensus/consensus.md
hanssv: So I guess I failed then _"I tried to make it obvious this isn't around until IRIS"_

But in a sense it isn't available in Iris, the next definition applies then...

consensus/consensus.md
tolbrino: That is why I approved nonetheless.

hanssv: 😅

PR #451, opened 2020-01-14
Virtual State Channels whitepaper

velzevur:
This is according to what had been discussed and described more informally in [the first proposition](https://github.com/aeternity/protocol/compare/VSC_proposal)

### DIFF
--- a/channels/virtual-state-channels.md
+++ b/channels/virtual-state-channels.md
@@ -1,0 +1,475 @@
+Virtual State Channels
+
+# Abstract
+
+The Aeternity blockchain provides various means for scaling the transaction
+throughput so it can accommodate billions of people. One of those is State
+Channels. This document describes how State Channels protocol can be further
+improved in order to build Virtual State Channels.
+
+# Introduction
+
+Aeternity blockchain already has two-party State Channels built on a protocol
+level. They allow producing off-chain transactions. They allow not just
+exchanging tokens off-chain but also executing smart contracts. That results
+in them not consuming neither `gas`, nor `fee` for their exectution. Off-chain
+transactions also don't need on-chain confirmations so their speed is bound by
+participant's computing power and network bandwidth. This allows building
+cheap near real-time protocols while keeping the same amount of trustlessness
+the blockchain already provides.
+
+While this is already a great improvement to traditional blockchain solutions,
+participants are still expected to interact with the on-chain for certain
+milestone events: opening a channel, depositing or withdrawing tokens and
+closing a channel. As with any on-chain transactions, the higher number of
+confirmations a transaction has, the better fork-safety it has. In the context
+of State Channels that results in big waiting times whenever the participants
+need to modify the on-chain. This is to be improved with the introduction of
+Virtual State Channels.
+
+State Channels, also known as on-chain State Channels, use the on-chain
+blockchain both as a source of truth and as an arbiter in disputes. Virtual
+State Channels build on top of them allowing to use a third party - an
+intermediary - as a source of truth or as an arbiter. It is crucial that, as
+with on-chain State Channels, a Virtual State Channel participant can protect
+themselves from malicious acts. Since third parties providing Virtual State
+Channel infrastructure can also act maliciously, at any point of time a
+participant can bring a potential dispute on-chain and have it resolved there.
+That way we keep Virtual State Channels as trustless as State Channels
+themselves.
+
+# Channels within channels
+
+On-chain State Channels have a state. It is represented as a MPTree and has
+the same structure as the on-chain state tree. A notable difference between
+those would be that the off-chain state tree is limited just to accounts and
+contracts. It does not allow having names, oracles or channels and the
+corresponding trees are empty. The proposed Virtual State Channels solution
+consists of allowing channels to be present in the off-chain state. This
+requires a corresponding dispute mechanism that protects all parties'
+interests and allows them to dispute the off-chain channel on-chain.
+
+Intention is instead of opening a State Channel on-chain, participants use an
+intermediary. Both participants must have an already opened on-chain State
+Channel with the intermediary. They open a Virtual State Channel within their
+on-chain State Channel's context and continue using it as they would an
+on-chain State Channel. Once opened, all off-chain Virtual State Channel
+communication goes directly from one participant to the other. The
+intermediary receives only Virtual State Channel updates only when required to
+take some action.
+
+With on-chain State Channels participants lock tokens in the channel itself. A
+similar approach is taken with Virtual State Channels: since participants are
+to create Virtual State Channel objects in the states of their on-chain State
+Channels with the intermediary, they all need to lock tokens in the new
+Virtual State Channel. They must lock the same amount of tokens in both
+on-chain State Channel states. A participant is representing herself in the
+Virtual State Channel and she locks as many tokens as she would like to. In
+the scope of the on-chain State Channel, the intermediary protects the
+interests of the other Virtual State Channel's participant and locks the
+corresponding amount of tokens from their behalf. This happens in both
+on-chain State Channel. As a result of this is the intermediary
+locks tokens in both on-chain State Channels and has an exposure as the total
+amount of tokens locked in the Virtual State Channel.
+
+#### Example
+
+```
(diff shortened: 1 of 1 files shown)


channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+throughput so it can accommodate billions of people. One of those is State
+Channels. This document describes how State Channels protocol can be further
+improved in order to build Virtual State Channels.
+
+## Introduction
+
+Aeternity blockchain already has two-party State Channels built on a protocol
+level. They allow producing off-chain transactions. They allow not just
+exchanging tokens off-chain but also executing smart contracts. That results
+in them not consuming neither `gas`, nor `fee` for their exectution. Off-chain
```
uwiger:
```suggestion
in them not consuming either `gas`, nor `fee` for their execution. Off-chain
```

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+transactions also don't need on-chain confirmations so their speed is bound by
+participant's computing power and network bandwidth. This allows building
```
uwiger:
```suggestion
the involved parties' computing power and network bandwidth. This allows building
```

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+cheap near real-time protocols while keeping the same amount of trustlessness
+the blockchain already provides.
+
+While this is already a great improvement to traditional blockchain solutions,
+participants are still expected to interact with the on-chain for certain
```
uwiger:
```suggestion
participants are still expected to interact with the chain for certain
```

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+milestone events: opening a channel, depositing or withdrawing tokens and
+closing a channel. As with any on-chain transactions, the higher number of
+confirmations a transaction has, the better fork-safety it has. In the context
+of State Channels that results in big waiting times whenever the participants
```
uwiger:
```suggestion
of State Channels that results in long waiting times whenever the participants
```

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+need to modify the on-chain. This is to be improved with the introduction of
+Virtual State Channels.
+
+State Channels, also known as on-chain State Channels, use the on-chain
+blockchain both as a source of truth and as an arbiter in disputes. Virtual
```
uwiger: "The on-chain blockchain" sounds funny

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+State Channels build on top of them allowing to use a third party - an
+intermediary - as a source of truth or as an arbiter. It is crucial that, as
+with on-chain State Channels, a Virtual State Channel participant can protect
```
uwiger:
```suggestion
with regular State Channels, Virtual State Channel participants can protect
```

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+themselves.
+
+## Channels within channels
+
+On-chain State Channels have a state. It is represented as a MPTree and has
+the same structure as the on-chain state tree. A notable difference between
+those would be that the off-chain state tree is limited just to accounts and
+contracts. It does not allow having names, oracles or channels and the
+corresponding trees are empty. The proposed Virtual State Channels solution
+consists of allowing channels to be present in the off-chain state. This
```
uwiger:
```suggestion
allows having channel objects stored in the off-chain state. This
```

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+requires a corresponding dispute mechanism that protects all parties'
+interests and allows them to dispute the off-chain channel on-chain.
+
+Intention is instead of opening a State Channel on-chain, participants use an
```
uwiger:
```suggestion
The intention is that instead of opening a State Channel on-chain, participants use an
```

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+intermediary. Both participants must have an already opened on-chain State
```
uwiger:
```suggestion
intermediary. Both participants must have an already opened State
```

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,500 @@
...
+Channel with the intermediary. They open a Virtual State Channel within their
+on-chain State Channel's context and continue using it as they would an
+on-chain State Channel. Once opened, all off-chain Virtual State Channel
+communication goes directly from one participant to the other. The
+intermediary receives only Virtual State Channel updates only when required to
```
uwiger:
```suggestion
intermediary receives Virtual State Channel updates only when required to
```

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,498 @@
...
+Virtual State Channel participants create and co-authenticate desired updates
+and provide them to the intermediary in the respective parent State Channels.
+In order for them to be applied there, the intermediary must also agree that
+those are valid updates. If the intermediary refuses to cooperate, most of
+those can be forcfully progressed on-chain. All of those will be explained in
+detail.
+
+A special case arises when both Virtual State Channel participants agree on an
+off-chain update and one of them provides it to the intermediary while the
+other does not. Since Channel state can only be updated with a mutual
```
uwiger: But would this happen? Didn't we say that the initiator of the update sends to the intermediary, and the intermediary forwards it to the other parent channel? This would be more similar to how it is on the chain (only one party posts to the chain, while the other gets notified of the on-chain change).

velzevur: Nope, it is the other way around: VSC participants create their updates and each of them provides them to the intermediary in their parent State Channels.

The other way around actually imposes a bit of a risk for the intermediary, namely:
1) Alice and Bob agree on stateX
2) Alice gives Ingrid the stateX
3) Ingrid gives Bob stateX but Bob refuses to update their VSC object

What shall Ingrid do now? In order for her to be safe, she must include that in their channel so she shall force progress it on-chain. On the other hand, Bob is in a possession of the update, already signed by Ingrid, and question is this gives him any advantage or not. If Ingrid ever authenticates an alternative update with the same `round`, she is in trouble.

A safer approach I am slightly in favour of would be both Alice and Bob provide her with the VSC off-chain update they have for her. If just one of them does it and the other does not, Alice could gently poke the non-cooperative one with a soft-force progress action or go directly on-chain forcing it there.

channels/virtual-state-channels.md
```diff
@@ -0,0 +1,498 @@
...
+participants, total balances dedicated to the channel and etc. It has all the
+required information to safely close the channel, even in cases of a dispute.
+
+#### Object structure
+
+In cases of Virtual State Channels we produce a regular State Channel object
+and wrap it around with some additional information. It consists of who the
+intermediary is and the initial token amounts dedicated by the participant and
+the intermediary. An important detail is that the initial token amounts for
+the participants are not according to the amounts they've dedicatrs: it is the
```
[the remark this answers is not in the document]
velzevur: Addressed

channels/virtual-state-channels.md
velzevur: I take that back. This is part of the category of problems _A proposes something authenticated to B, B refuses but is still in possession of the authenticated object_ . The worst thing could happen is Bob to force progress based on that particular state - hence approving it. The flow who tells who what should not really matter that much and we need propagation of messages in multy intermediary scenarios anyway.

The described approach is not mandatorary but prefered for the symmetry it provides.

uwiger: But the safer approach would mean that Ingrid gets an update from Alice, and then waits for the same tx to be provided by Bob, eventually perhaps poking with a soft-FP. If Ingrid instead performs the soft-FP right away, Bob would be waiting in much the same way as in the on-chain case, and refusing would either have the result that Alice FPs on the parent chain/channel or simply solo-closes and declares the VSC uncooperative.

radrow:
Is this still a thing?

PR #463, opened 2020-02-25
Reformatted and reworked stdlib documentation. Added builtins

radrow:
I added all builtin functions I know to the sophia stdlib doc. Also I reformatted it to make it display more nicely and make it easier to automatically reprocess by some script. Also fixed some overlooks.

### DIFF
--- a/contracts/sophia_stdlib.md
+++ b/contracts/sophia_stdlib.md
@@ -358,7 +358,7 @@ Oracle.extend(<signature : bytes(64)>, o : oracle('a, 'b), ttl : Chain.ttl) : un
 ```
 
 Extends TTL of an oracle.
-* `singature` is a named argument and thus optional. Must be the same as for `Oracle.register`
+* `signature` is a named argument and thus optional. Must be the same as for `Oracle.register`
 * `o` is the oracle being extended
 * `ttl` must be `RelativeTTL`. The time to live of `o` will be extended by this value.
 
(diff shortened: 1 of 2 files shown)


contracts/sophia_stdlib.md
```diff
@@ -1,95 +1,711 @@
...
+
+
+## Address
+
+### to_str
+```
+Address.to_str(a : address) : string
+```
+
+Base58 encoded string
```
hanssv:
```suggestion
Base58Check encoded string
```

contracts/sophia_stdlib.md
```diff
@@ -1,95 +1,711 @@
...
+signing the oracle query id + contract address
+
+
+### extend
+```
+Oracle.extend(<signature : bytes(64)>, o : oracle('a, 'b), ttl : Chain.ttl) : unit
+```
+
+Extends TTL of an oracle.
+* `singature` is a named argument and thus optional. Must be the same as for `Oracle.register`
```
hanssv:
```suggestion
* `signature` is a named argument and thus optional. Must be the same as for `Oracle.register`
```

radrow: Good catch!

tolbrino:
@radrow Isn't this PR replaced by https://github.com/aeternity/aesophia/pull/239 ?

radrow:
Yes, it is

PR #488, opened 2021-06-08
Mention factories opcodes, remove varna

radrow:


radrow:
The formatting in the md file is not as nice as it used to be, because I autogenerated it and didn't want to bother to write a padder

cytadela8 (approved):

cytadela8:
Great! Could we get also some explenation of the CLONE opcode somewhere. I think it's behaviour may be non-obvious.

cytadela8 (approved):

PR #494, opened 2021-07-31
docs: restructuring & introduction of mkdocs

marc0olo:
the current state can be viewed here: https://marc0olo.github.io/protocol/

I made some restructuring and added mkdocs to finally be able to find what somebody needs more easily.

I also noticed parts that definitely need to be reviewed again as they aren't really accurate any longer. but this should be done separately IMO.

this PR focuses on providing a guided and clickable structure for the protocol.

when this is merged a github action should automatically deploy the docs which should then be accesible at https://aeternity.com/protocol/


dincho:
What's the intention behind moving everything in a subdir ? While I agree it might look better from code perspective, unfortunately this is major BC break and I bet a lot of back-links will be broken after that. Also there are 14 pull requests here that this will create conflicts for.

So I'd suggest keeping the old file location for now at least.

mkdocs-hook.py
```diff
@@ -0,0 +1,4 @@
+import shutil
```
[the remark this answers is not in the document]
marc0olo: will check that

marc0olo:
> What's the intention behind moving everything in a subdir ? While I agree it might look better from code perspective, unfortunately this is major BC break and I bet a lot of back-links will be broken after that. Also there are 14 pull requests here that this will create conflicts for.
> 
> So I'd suggest keeping the old file location for now at least.

I generally agree. not 100% sure anymore why I did that. will try to handle this via the simple-hook.

independently I changed some structure and renamed index files in subfolders to `README.md`. should we also try to keep the filenames wherever possible? just wondering

mkdocs-hook.py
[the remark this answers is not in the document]
marc0olo: ok I remember now. at first I wanted to include it in the `.github/workflows` folder but somehow the path with the dot made problems there and then I decided to place it under root.

what about a `scripts` folder?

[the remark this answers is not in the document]
marc0olo: I now have a specific mkdocs folder `00_mkdocs` where everything related to it is stored and being copied to via the hook

marc0olo:
now using the old structure again. there are still some changes (e.g. `sync` folder and the renaming of the entrypoint markdowns to `README.md`)

can we keep that or should I revert that, too?

marc0olo:
improved the docs structure. using a separate `.docssite` folder now for the mkdocs stuff

dincho:
I'd suggest keeping the old file name as well and point it with link to the new index file (aka README.md)

AENS.md
```diff
@@ -3,7 +3,7 @@
 
 ## Abstract
 
-This document describes the first draft of the Aeternity Naming System,
+This document describes the first draft of the AEternity Naming System,
```
dincho, line 6: wut ?

channels/OFF-CHAIN.md
```diff
@@ -853,7 +853,7 @@ useable.
 Message code: 18
 
 Since both sides may initiate a `withdraw_created` request, it is possible
-that both may do so at roughly the same time. In particular, in the Aeternity node
```
dincho, line 856: I think there was a discussion about capitalisation, I'm definitely not an English expert, but did you double checked it should be lowercase ?

contracts/events.md
```diff
@@ -12,7 +11,7 @@ The events are stored in the [contract call object](../serializations.md#contrac
 and can be accessed by looking up the transaction in which they were emitted. 
 
 Let's consider the following Sophia example:
-```
+```sophia
```
dincho, line 14: I doubt github highlighter supports that, if you find a closer to syntax language that would be better isn't it ? If not, this is better than missing annotation I guess.

marc0olo: well I hope I can push this topic forward. there is a quite old PR about this. so I wanted at least to prepare the code, see https://github.com/github/linguist/pull/4719.

we need to revisit this.

channels/OFF-CHAIN.md
marc0olo: to be discussed. maybe this needs to be addressed in another PR. we should be consistent here

node/api/README.md
```diff
@@ -235,15 +247,4 @@ The `'data'` element may contain more detailed information inside `'data'`:
...
-
-* [Encoding scheme for API identifiers](./api_encoding.md)
-* [Account management user API usage](./account_api_usage.md)
-* [Spending tokens using user API](./spend_api_usage.md)
-* [Oracle user API usage](./oracle_api_usage.md)
-* [Naming system API usage](./naming_system_api_usage.md)
-* [Contract API usage](./contract_api_usage.md)
-* [Channels API usage](./channels_api_usage.md)
-* [Mining API usage](./mining_api_usage.md)
+Detailed message transcripts from test suites can be found [here](https://github.com/aeternity/protocol/tree/master/specification/node/api/examples/channels/json-rpc).
```
dincho, line 250: why absolute URL ?

channels/OFF-CHAIN.md
dincho: any update on this ? consistency can be accompanied by `aeternity` -> `Aeternity` as well.

AENS.md
dincho: I don't get it ? You want to emphasise on abbreviation? Even that does not makes much sense to me :)

node/api/README.md
marc0olo: will be fixed :D

AENS.md
marc0olo: ahh I think this was updated. should be fixed already. but I will double check.

channels/OFF-CHAIN.md
marc0olo: seems like I missed that one. will check again. thanks

velzevur (approved):

dincho (approved):

PR #526, opened 2024-03-07
Add ping version 2 description.

mitchelli:
This PR is supported by Æternity Foundation and Æternity Crypto Foundation

### DIFF
--- a/sync/p2p_messages.md
+++ b/sync/p2p_messages.md
@@ -80,14 +80,33 @@ Message is RLP encoded, fields:
 
 Message is RLP encoded, fields:
 
-  - `Port :: int` - listen port
-  - `Share :: int` - number of peers to share
+### Version 1
+
+  - `Port :: int` - listen port.
+  - `Share :: int` - number of peers to share.
   - `GenesisHash :: byte_array`
   - `Difficulty :: int` - the total difficulty of the chain.
   - `TopHash :: byte_array`
-  - `sync_allowed :: bool` - if the sender of this ping message is accepting
+  - `SyncAllowed :: bool` - if the sender of this ping message is accepting
     synchronization messages.
-  - `Peers :: [byte_array]` - list of shared peers
+  - `Peers :: [byte_array]` - list of shared peers.
+
+
+### Version 2
+
+  - `Versions :: list`
+    - `Protocol :: binary` - type of message.
+    - `Vsns :: [int]` - versions supported
+  - `Port :: int` - listen port.
+  - `Share :: int` - number of peers to share.
+  - `GenesisHash :: byte_array`
+  - `Height :: int` - current height.
+  - `Difficulty :: int` - the total difficulty of the chain.
+  - `TopHash :: byte_array`
+  - `SyncAllowed :: bool` - if the sender of this ping message is accepting
+    synchronization messages.
+  - `Capabilities :: byte_array` - list of supported capabilites.
+  - `Peers :: [byte_array]` - list of shared peers.
 
 Peers are serialized/deserialized in `aec_peer_messages`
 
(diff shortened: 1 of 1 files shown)


sync/p2p_messages.md
```diff
@@ -80,14 +80,33 @@ Message is RLP encoded, fields:
...
+  - `TopHash :: byte_array`
+  - `SyncAllowed :: bool` - if the sender of this ping message is accepting
+    synchronization messages.
+  - `Peers :: [byte_array]` - list of shared peers.
+
+
+### Version 2
+
+  - `Versions :: list`
+    - `Protocol :: binary` - type of message.
```
dincho, line 98: Could you please elaborate a bit more on this field? It's name is protocol, it's description is "type of message".

In general I'd like to see somewhere documented how this "Versions" field should work and what is used for.

sync/p2p_messages.md
```diff
@@ -80,14 +80,33 @@ Message is RLP encoded, fields:
...
+    - `Vsns :: [int]` - versions supported
+  - `Port :: int` - listen port.
+  - `Share :: int` - number of peers to share.
   - `GenesisHash :: byte_array`
+  - `Height :: int` - current height.
   - `Difficulty :: int` - the total difficulty of the chain.
   - `TopHash :: byte_array`
-  - `sync_allowed :: bool` - if the sender of this ping message is accepting
+  - `SyncAllowed :: bool` - if the sender of this ping message is accepting
```
dincho, line 106: Since we have a capability field now, isn't it appropriate to move this there? Sounds like a capability to me.

sync/p2p_messages.md
```diff
@@ -80,14 +80,33 @@ Message is RLP encoded, fields:
...
+
+### Version 2
+
+  - `Versions :: list`
+    - `Protocol :: binary` - type of message.
+    - `Vsns :: [int]` - versions supported
+  - `Port :: int` - listen port.
+  - `Share :: int` - number of peers to share.
   - `GenesisHash :: byte_array`
+  - `Height :: int` - current height.
```
dincho, line 103: Isn't height kind of "per database" (node) field? Why is it shared? If you know the remote TopHash isn't it possible to derive its height as well?

sync/p2p_messages.md
mitchelli: At the moment, the only protocol is ping. It specifies the node supports v1 and v2 pings. It will be used to decide which version of a message should be used. I'll add an explanation.

sync/p2p_messages.md
mitchelli: It could be, the current example capability is more complicated though, it is for a sync capability "chain_poi", which is:
```
{height, int}
{root_hash, binary}
{genesis, binary}
{top, binary}
{poi, binary} 
```
A downside would be that it increases the size of the ping.

dincho: Okay, maybe a misunderstand "capabilities"

sync/p2p_messages.md
mitchelli: I asked Ulf and this is his response, "It's true that the height can be derived, but only by first downloading headers. While working on fast synch strategies, I felt that getting the height up front was a convenience at fairly low cost."

sync/p2p_messages.md
hanssv: I don't think we are sure that is how capabilities should be used, let's see when we actually find a use-case for them. My idea was that the capabilities should signal what _additional_ messages that the node will understand and react to 🤔

hanssv (approved):

ThomasArts (approved):

sync/p2p_messages.md
mitchelli: Ulf wanted to use it for additional methods of sync. https://github.com/aeternity/aeternity/issues/4189

